PR #163, opened 2026-04-21
chore: fix fuzz-found parser safety issues (stack overflow + OOM)

usercise:
## Summary

- **Stack overflow** (`fix: stack overflow from unbounded recursion in walk_node`, already committed): `walk_node` and `find_identifier` had no depth limit; malformed tree-sitter ASTs from binary fuzz input caused unbounded recursion. Fixed by adding `MAX_WALK_DEPTH = 512` and `MAX_IDENT_DEPTH = 64` guards with a `depth` parameter threaded through both functions.

- **Out-of-memory** (this commit): a single ~3.8 KB adversarial input caused RSS to spike from 541 MB to 2 GB in one `SourceParser::parse` call. Two guards close the window:
  - **Parse timeout** (5 s via `ParseOptions::progress_callback`): aborts tree-sitter's GLR phase if it overruns and falls back to sliding-window chunking, preventing the parser from allocating unbounded memory on ambiguous input.
  - **Chunk cap** (`MAX_CHUNKS = 100 000` in `walk_node`): returns early if the output `Vec` reaches the limit, bounding memory for pathologically wide parse trees.

All three fuzz artifacts (`crash-*` × 3, `oom-*` × 2) are addressed by these changes.

## Test plan

- [ ] `cargo test` passes (all suites green)
- [ ] `cargo build --release` succeeds
- [ ] Re-run fuzzer against the artifact inputs: `cargo +nightly fuzz run fuzz_parser fuzz/artifacts/fuzz_parser/` — no crashes or OOMs

### DIFF
--- a/src/indexer/parser/ts_walker.rs
+++ b/src/indexer/parser/ts_walker.rs
@@ -142,7 +142,7 @@ pub(super) fn node_specs(language: &str) -> Vec<NodeSpec> {
 }
 
 /// Maximum AST recursion depth.  Deeply-nested or pathological parse trees
-/// (common with fuzz inputs) would otherwise overflow the stack.
+/// (common with adversarial inputs) would otherwise overflow the stack.
 const MAX_WALK_DEPTH: usize = 512;
 
 /// Maximum number of chunks collected in a single walk.  A file with millions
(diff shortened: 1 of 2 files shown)


src/indexer/parser/ts_walker.rs
```diff
@@ -141,6 +141,15 @@ pub(super) fn node_specs(language: &str) -> Vec<NodeSpec> {
     }
 }
 
+/// Maximum AST recursion depth.  Deeply-nested or pathological parse trees
+/// (common with fuzz inputs) would otherwise overflow the stack.
```
usercise:
```suggestion
/// (common with adversarial inputs) would otherwise overflow the stack.
```

PR #235, opened 2026-05-17
bench(memory): give grep baseline a fair query in Decision Archaeology

usercise:
Closes #227

Replace single rigged grep_literal condition with four:
- grep_literal: git log --grep with full question (labelled honestly)
- grep_keywords: regex-extract keywords, grep each, merge results
- fts_commit_messages: SQLite FTS5 index over all commits
- memory_search: unchanged spelunk memory search

All four run on the same question set. Output JSON includes recall, MRR, and median wall per condition. Hit checking matches ground_truth_commit SHA.

### DIFF
--- a/bench/memory/decision_archaeology.py
+++ b/bench/memory/decision_archaeology.py
@@ -31,6 +31,7 @@ import re
 import sqlite3
 import statistics
 import subprocess
+import sys
 import time
 from datetime import datetime, timezone
 from pathlib import Path
@@ -49,21 +50,15 @@ def extract_keywords(question: str) -> list[str]:
     """
     keywords: list[str] = []
 
-    # Quoted strings
     quoted = re.findall(r'"([^"]+)"', question)
     keywords.extend(quoted)
-
-    # CAPITALISED acronyms
     caps = re.findall(r"\b[A-Z][A-Z_]{2,}\b", question)
     keywords.extend(caps)
-
-    # snake_case and CamelCase identifiers
     snake = re.findall(r"\b[a-z]+_[a-z_]+\b", question, re.IGNORECASE)
     keywords.extend(snake)
     camel = re.findall(r"\b[A-Z][a-z]+(?:[A-Z][a-z]+)+\b", question)
     keywords.extend(camel)
 
-    # Deduplicate, sort by length desc, take top 5
     seen = set()
     unique = []
     for kw in keywords:
@@ -136,12 +131,10 @@ def _git_log_grep(repo_path: Path, pattern: str, limit: int = 10) -> list[dict]:
 
 
 def run_grep_literal(repo_path: Path, query: str, limit: int = 10) -> list[dict]:
-    """Grep using the full question string (verbatim)."""
     return _git_log_grep(repo_path, query, limit)
 
 
 def run_grep_keywords(repo_path: Path, query: str, limit: int = 10) -> list[dict]:
-    """Extract keywords from the question, grep for each, merge results."""
     keywords = extract_keywords(query)
     all_results: list[dict] = []
     seen_commits: set[str] = set()
@@ -156,7 +149,6 @@ def run_grep_keywords(repo_path: Path, query: str, limit: int = 10) -> list[dict
 
 
 def _build_fts_index(repo_path: Path, rebuild: bool = False) -> Path:
-    """Build a SQLite FTS5 index over commit messages. Returns path to the DB."""
     db_path = repo_path / ".git" / "spelunk_fts_commits.db"
     if rebuild and db_path.exists():
         db_path.unlink()
@@ -169,7 +161,6 @@ def _build_fts_index(repo_path: Path, rebuild: bool = False) -> Path:
     )
     conn.commit()
 
-    # Stream commits via git log
     proc = subprocess.Popen(
         ["git", "--no-pager", "log", "--format=%H%x00%s%x00%b%x00---"],
         cwd=repo_path,
@@ -197,18 +188,25 @@ def _build_fts_index(repo_path: Path, rebuild: bool = False) -> Path:
     return db_path
 
 
-def run_fts_commit_messages(repo_path: Path, query: str, limit: int = 10) -> list[dict]:
+def run_fts_commit_messages(
+    repo_path: Path, query: str, limit: int = 10, rebuild_fts: bool = False
+) -> list[dict]:
     """Build FTS5 index over commit messages and query with the full question."""
     try:
-        db_path = _build_fts_index(repo_path)
+        db_path = _build_fts_index(repo_path, rebuild=rebuild_fts)
         conn = sqlite3.connect(str(db_path))
         rows = conn.execute(
-            "SELECT commit, title, body FROM commits WHERE commits MATCH ? ORDER BY rank LIMIT ?",
+            "SELECT sha, title, body FROM commits WHERE commits MATCH ? ORDER BY rank LIMIT ?",
             (query, limit),
         ).fetchall()
         conn.close()
         return [{"commit": r[0], "title": r[1], "body": r[2]} for r in rows]
-    except Exception:
+    except sqlite3.OperationalError as e:
+        # FTS query syntax errors — treat as no results
+        print(f"  FTS query failed (syntax): {e}", file=sys.stderr)
+        return []
+    except Exception as e:
+        print(f"  FTS query failed: {type(e).__name__}: {e}", file=sys.stderr)
         return []
 
 
@@ -276,7 +274,6 @@ def main() -> None:
     print(f"Questions: {len(questions)}")
     print()
 
-    # Accumulators per condition
     conditions = {
         "grep_literal": {"hits": [], "ranks": [], "wall": []},
         "grep_keywords": {"hits": [], "ranks": [], "wall": []},
@@ -314,7 +311,6 @@ def main() -> None:
             status = "HIT" if hit else "MISS"
             print(f"  {cond_name:<22} {status:4} (rank={rank or '-'}, {elapsed:.2f}s)")
 
-    # Build output
     output: dict = {
         "benchmark": "decision_archaeology",
         "repo": str(repo_path),
--- a/bench/memory/decision_archaeology.py
+++ b/bench/memory/decision_archaeology.py
@@ -188,16 +188,26 @@ def _build_fts_index(repo_path: Path, rebuild: bool = False) -> Path:
     return db_path
 
 
+def _sanitize_fts_query(query: str) -> str:
+    """Strip FTS5-special characters; return space-separated tokens."""
+    cleaned = "".join(c if c.isalnum() or c.isspace() else " " for c in query)
+    tokens = cleaned.split()
+    return " ".join(tokens) if tokens else ""
+
+
 def run_fts_commit_messages(
     repo_path: Path, query: str, limit: int = 10, rebuild_fts: bool = False
 ) -> list[dict]:
     """Build FTS5 index over commit messages and query with the full question."""
     try:
+        clean = _sanitize_fts_query(query)
+        if not clean:
+            return []
         db_path = _build_fts_index(repo_path, rebuild=rebuild_fts)
         conn = sqlite3.connect(str(db_path))
         rows = conn.execute(
             "SELECT sha, title, body FROM commits WHERE commits MATCH ? ORDER BY rank LIMIT ?",
-            (query, limit),
+            (clean, limit),
         ).fetchall()
         conn.close()
         return [{"commit": r[0], "title": r[1], "body": r[2]} for r in rows]
--- a/bench/memory/decision_archaeology.py
+++ b/bench/memory/decision_archaeology.py
@@ -1,44 +1,86 @@
 #!/usr/bin/env python3
 """Decision archaeology benchmark — measure spelunk memory retrieval vs grep.
 
-Evaluates whether `spelunk memory search` can retrieve decisions, context,
-and design rationale from git history better than raw `git log --grep`.
+Four conditions compared against the same question set:
 
-Workflow:
-  1. Index the repo:           spelunk index <repo>
-  2. Harvest memory:           spelunk memory harvest --git-range <range>
-  3. Run this benchmark:       python bench/memory/decision_archaeology.py ...
-
-For each curated question, runs both memory search and grep baseline,
-checking whether ground-truth keywords or commits appear in results.
+    grep_literal    — git log --grep with the full question string (verbatim)
+    grep_keywords   — git log --grep with regex-extracted keywords from question
+    fts_commit_msgs — SQLite FTS5 index over all commit messages, full question
+    memory_search   — spelunk memory search (semantic over harvested entries)
 
 Usage:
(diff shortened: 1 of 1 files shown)


usercise (commented):
**Status: changes needed (one blocker, three nits)**

Thanks for tackling this — the four-condition structure is exactly what #227 asked for, and `grep_literal` being honestly labelled is a real win. One blocker and three smaller concerns before this can merge, in priority order.

## Blocker — `memory_search` will score 0% on every run

The new `check_hit` in `bench/memory/decision_archaeology.py:214-218` matches the ground-truth commit against `r.get("commit", "")`:

```python
if commit and r.get("commit", "") == commit:
    return True, i + 1
```

This works for `grep_literal`, `grep_keywords`, and `fts_commit_messages` — those return records with a `commit` field. But `spelunk memory search --format json` returns memory entries where the SHA lives under `source_ref`, not `commit`:

```json
{
  "id": 1,
  "kind": "decision",
  "title": "bleat a DEBUG message when RIPGREP_CONFIG_PATH is not set",
  "source_ref": "cb66736f146f093497f4dc537b33d0826f9af33c",
  ...
}
```

So `r.get("commit", "")` returns `""` for every memory entry → recall=0, MRR=0 regardless of how well memory actually retrieved.

Compounding issue: `questions-ripgrep.json` uses short SHAs (`"cb66736f"`) while `source_ref` is the full 40-char SHA, so even if you fixed the field name, an `==` comparison would still miss. The hit check needs to handle both fields and prefix-match:

```python
def check_hit(results: list[dict], commit: str) -> tuple[bool, int | None]:
    if not commit:
        return False, None
    for i, r in enumerate(results):
        candidate = r.get("commit") or r.get("source_ref") or ""
        if candidate.startswith(commit) or commit.startswith(candidate):
            return True, i + 1
    return False, None
```

Without this, the benchmark flips the previous \"rigged in spelunk's favour\" problem into a \"rigged against spelunk\" problem.

## Smaller concerns

**FTS index is permanently cached** — `_build_fts_index` at `decision_archaeology.py:152-156` does `if db_path.exists(): return db_path`. Re-running after new commits land won't reindex, so different runs of the \"same\" benchmark silently compare current memory against a stale FTS snapshot. Suggest either rebuilding on every run or adding a `--rebuild-fts` flag (and noting the cache strategy in the result JSON).

**`commit` is a SQL keyword** — `CREATE VIRTUAL TABLE commits USING fts5(commit, title, body)`. It happens to work in current SQLite but it's a footgun. Rename the column to `sha` (or quote `\"commit\"` consistently in DDL+DML).

**Keyword extractor falls back to `question.split()[0]`** — `decision_archaeology.py:92`. For plain-English questions like \"How does error handling work in the parser?\", none of the regex patterns (CAPS / snake_case / CamelCase / quotes) match, and the fallback grep-for-\"How\" returns garbage. Either let `grep_keywords` legitimately return `[]` and score 0 honestly, or use a richer extractor (stop-word strip + remaining content words).

## What's good

- Four-condition structure is exactly the shape #227 asked for.
- `_git_log_grep` is cleanly factored out and reused by both grep conditions.
- Per-condition recall / MRR / median_wall in the output JSON, plus the per-question `details` array.
- FTS schema is fine modulo the column name; streaming insert with batch size 1000 is appropriate.

Once the `check_hit` fix lands, this is good to go.

usercise (commented):
**Status: original blocker is fixed, but the fixup introduces two new bugs that silently make `fts_commit_messages` score 0% on every question.**

The good news first — `check_hit` now correctly handles both `commit` and `source_ref` with prefix matching, the keyword extractor returns `[]` honestly instead of falling back to a stop word, and the `--rebuild-fts` flag + `fts_rebuilt` in the output JSON are exactly right. The original blocker on memory_search hit detection is resolved.

But the FTS column rename and the rebuild-flag plumbing didn't go all the way through. Two issues, both silently swallowed by the bare `except Exception: return []` at `run_fts_commit_messages`:

## Bug 1 — SELECT still references the old column name

`decision_archaeology.py:206`:

```python
rows = conn.execute(
    "SELECT commit, title, body FROM commits WHERE commits MATCH ? ORDER BY rank LIMIT ?",
    (query, limit),
).fetchall()
```

The DDL and INSERT were updated to use `sha` (good), but this SELECT was missed. At runtime SQLite will raise `OperationalError: no such column: commit`, which the surrounding `try/except Exception: return []` catches → `fts_commit_messages` returns empty every time.

Fix:

```python
rows = conn.execute(
    "SELECT sha, title, body FROM commits WHERE commits MATCH ? ORDER BY rank LIMIT ?",
    (query, limit),
).fetchall()
return [{"commit": r[0], "title": r[1], "body": r[2]} for r in rows]
```

(Keep the result dict key as `commit` so `check_hit` finds it — only the FTS column name needs renaming.)

## Bug 2 — `run_fts_commit_messages` doesn't accept `rebuild_fts`

`decision_archaeology.py:200`:

```python
def run_fts_commit_messages(repo_path: Path, query: str, limit: int = 10) -> list[dict]:
```

But the call site at `decision_archaeology.py:300-302` now passes:

```python
results = run_fts_commit_messages(
    repo_path, question, args.limit, rebuild_fts=args.rebuild_fts
)
```

This raises `TypeError: run_fts_commit_messages() got an unexpected keyword argument 'rebuild_fts'`, also caught silently. Even when `--rebuild-fts` isn't passed (default `False`) the TypeError still fires because Python evaluates the kwarg unconditionally.

Fix:

```python
def run_fts_commit_messages(
    repo_path: Path, query: str, limit: int = 10, rebuild_fts: bool = False
) -> list[dict]:
    try:
        db_path = _build_fts_index(repo_path, rebuild=rebuild_fts)
        ...
```

## Suggestion — narrow the `except Exception`

These two bugs would have surfaced immediately if the FTS function logged the exception. The bare `except Exception: return []` is hiding real failures and making the benchmark's `recall=0` numbers indistinguishable from \"actual zero results\". At minimum:

```python
except Exception as e:
    print(f"FTS query failed: {e}", file=sys.stderr)
    return []
```

Or split: catch `sqlite3.OperationalError` (treat as no-results) but let `TypeError` propagate. Without this, the next bug in this code path also goes unnoticed.

## Sanity check before merging

After the fixes land, a quick `python bench/memory/decision_archaeology.py --repo-path /Users/johan/opensource/ripgrep --questions bench/memory/questions-ripgrep.json` should report non-zero numbers for *both* `memory_search` and `fts_commit_messages`. If either is still 0, there's another bug. That's the smoke test I'd want green before approving.

Once these are fixed, this is good to merge.

usercise (commented):
**Status: two prior bugs fixed, one new bug found by the new stderr logging.**

Pulled the branch and ran the smoke test against ripgrep with the existing question set:

```
grep_literal           recall=0.00%   MRR=0.000  wall=0.040s
grep_keywords          recall=20.00%  MRR=0.200  wall=0.000s
fts_commit_messages    recall=0.00%   MRR=0.000  wall=0.036s
memory_search          recall=100.00% MRR=0.833  wall=0.040s
```

✅ **`memory_search`** is now hitting correctly (5/5 questions, MRR=0.833) — the `source_ref` + prefix-match fix in `check_hit` works.
✅ **`grep_keywords`** behaves honestly — caught `RIPGREP_CONFIG_PATH` on Q1, returned no results on the plain-English Q5 (\"What version was ripgrep updated to recently?\") because no technical tokens were extractable. That's the intended behaviour.
✅ The split exception handling is doing its job — the new bug below was visible in stderr immediately.

## New blocker — FTS5 query parser chokes on `?`

stderr from the run:

```
FTS query failed (syntax): fts5: syntax error near "?"
FTS query failed (syntax): fts5: syntax error near "?"
FTS query failed (syntax): fts5: syntax error near "?"
FTS query failed (syntax): fts5: syntax error near "?"
FTS query failed (syntax): fts5: syntax error near "?"
```

Every question in `questions-ripgrep.json` ends in `?`. FTS5's MATCH parser treats `?` as a special character (not a valid operator), so the whole query fails to parse. None of the FTS queries are actually executing. Same problem would hit any natural-language question with `?`, `.`, `:`, apostrophes, etc.

Fix: sanitize the query before passing to FTS5. Simplest version — strip non-alphanumeric characters, collapse whitespace:

```python
def _sanitize_fts_query(query: str) -> str:
    # Replace FTS5-special chars with spaces, collapse whitespace
    cleaned = "".join(c if c.isalnum() or c.isspace() else " " for c in query)
    tokens = cleaned.split()
    if not tokens:
        return ""
    return " ".join(tokens)


def run_fts_commit_messages(
    repo_path: Path, query: str, limit: int = 10, rebuild_fts: bool = False
) -> list[dict]:
    try:
        clean = _sanitize_fts_query(query)
        if not clean:
            return []
        db_path = _build_fts_index(repo_path, rebuild=rebuild_fts)
        conn = sqlite3.connect(str(db_path))
        rows = conn.execute(
            "SELECT sha, title, body FROM commits WHERE commits MATCH ? ORDER BY rank LIMIT ?",
            (clean, limit),
        ).fetchall()
        ...
```

After this fix, Q1 (`Why was RIPGREP_CONFIG_PATH added?` → `Why was RIPGREP_CONFIG_PATH added`) should match the `cb66736 core: bleat a DEBUG message when RIPGREP_CONFIG_PATH is not set` commit, getting `fts_commit_messages` off zero.

Note that the FTS5 default tokenizer (`unicode61`) splits on punctuation including underscores, so `RIPGREP_CONFIG_PATH` is indexed as three tokens (`RIPGREP`, `CONFIG`, `PATH`). The sanitised query produces the same three tokens. That should be fine.

## Caveat on the headline number

The 100% memory vs 0% grep result is still on the **corpus-derived** question set (`questions-ripgrep.json` was authored by reading harvested memory entries). The real test happens after [#237](https://github.com/usercise/spelunk/issues/237) lands and we have blind question sets. For now, the *purpose* of this smoke test is just to confirm all four conditions execute correctly — three out of four do. Fix the FTS sanitisation and we're at four out of four.

Otherwise the code is in good shape. Two iterations is the right number for this — the structural design is solid, the bugs were all in edge handling.

usercise (commented):
**Status: sanitiser works (no more syntax errors), but FTS5 still scores 0% — one more layer needed.**

The sanitiser successfully strips `?` and other FTS5-special chars, so the SQL no longer raises. But FTS recall is still 0% across all five questions. Investigated with the FTS db directly:

```python
# Sanitised Q1: "Why was RIPGREP_CONFIG_PATH added"
SELECT ... FROM commits WHERE commits MATCH 'Why was RIPGREP_CONFIG_PATH added'
# → []
SELECT ... FROM commits WHERE commits MATCH 'RIPGREP_CONFIG_PATH'
# → 2 rows including the target cb66736
```

FTS5 defaults to **AND** semantics across whitespace-separated tokens — the multi-word query requires every one of `Why`, `was`, `RIPGREP_CONFIG_PATH`, `added` to appear in the same commit message. For natural-language questions that's never going to happen — commit subjects don't repeat the phrasing of the questions a developer asks about them.

This is the same shape of problem the original `grep_literal` had with `git log --grep`, just one layer deeper. Without the AND→OR conversion, `fts_commit_messages` is essentially a fancier version of `grep_literal`.

## Fix

OR-join the sanitised tokens, with each token quoted (so FTS5 reserved keywords like `AND`/`OR`/`NEAR` in the question don't get re-interpreted as operators):

```python
FTS5_RESERVED = {"and", "or", "not", "near"}

def _sanitize_fts_query(query: str) -> str:
    """Convert a natural-language query to an FTS5 OR-of-quoted-tokens query."""
    cleaned = "".join(c if c.isalnum() or c.isspace() else " " for c in query)
    tokens = [
        t for t in cleaned.split()
        if len(t) >= 2 and t.lower() not in FTS5_RESERVED
    ]
    return " OR ".join(f'"{t}"' for t in tokens) if tokens else ""
```

Length filter (`>= 2`) drops one-char noise. FTS5 reserved word filter prevents accidental operator interpretation. Quoting each token also protects against tokens that contain characters FTS5 might still mishandle.

## Verified recall after the fix

Ran the OR-joined query manually against the existing ripgrep FTS db, top-10 results:

| Q | Target | Rank |
|---|---|---|
| Why was RIPGREP_CONFIG_PATH added? | `cb66736f` | **5** |
| What container types were added for Docker support? | `9b84e154` | **5** |
| What change was made to path checking in ignore logic? | `85edf4c7` | — |
| How was command existence verification updated? | `0a88cccd` | — |
| What version was ripgrep updated to recently? | `af60c2de` | — |

After the OR fix: `fts_commit_messages` recall@10 = 40%, MRR ≈ 0.2 — alongside `memory_search` at 100% / MRR 0.833 and `grep_keywords` at 20%. All four conditions producing signal.

The three remaining misses (Q3–Q5) are *honest* — the commit messages share no salient terms with the question wording (e.g. Q5's commit is literally just `\"15.1.0\"`, no overlap with \"What version was ripgrep updated to recently?\"). Memory wins those because the harvested entry paraphrases the commit. That's the differentiation the benchmark is meant to expose.

## Side note

This is the right time to remind us that the 100% / 40% / 20% / 0% pattern above is still on the **corpus-derived** question set. [#237](https://github.com/usercise/spelunk/issues/237) (blind question authoring) is still the gate before any of these numbers are publishable. But once #237 lands, all four conditions will at least be able to compete honestly.

Iteration count is fine — the structural design has been right since iteration 1; we're just finishing the long tail of natural-language-vs-FTS5 corner cases.

usercise (commented):
**Status: ✅ all four conditions now produce signal. Ready to merge from my side (would be `--approve` but GitHub won't let me approve my own org's PR).**

Pulled `008cbd0` and re-ran the smoke test against ripgrep:

```
grep_literal           recall=0.00%   MRR=0.000  wall=0.040s
grep_keywords          recall=20.00%  MRR=0.200  wall=0.000s
fts_commit_messages    recall=40.00%  MRR=0.080  wall=0.038s
memory_search          recall=100.00% MRR=0.833  wall=0.027s
```

Stderr is empty — no exceptions silently swallowed.

The shape is what we wanted: semantic > FTS > keyword > literal. FTS's MRR (0.080) is correctly lower than its recall (0.40) because the target commits land mid-list (rank ~5) due to OR-query noise from common words — honest BM25 ranking, not a bug. Memory wins because the harvested entries paraphrase the commits and bridge the natural-language → commit-message gap that lexical search can't cross.

`FTS5_RESERVED` filter and the `len(t) >= 2` guard are exactly the right shape for keeping the query safe against future questions containing FTS5 operator words.

## Reminder for the rerun

These numbers are still on the **corpus-derived** `questions-ripgrep.json` (5 questions authored by reading harvested memory). The point of this smoke test was structural — confirm the four conditions execute and produce sensible numbers — not to interpret the result. The real run that produces publishable numbers will use the blind question sets from [#237](https://github.com/usercise/spelunk/issues/237) (≥30 questions × ≥3 repos), which is now the gating issue for the Decision Archaeology section of [#232](https://github.com/usercise/spelunk/issues/232) (report rewrite).

Solid work across the four iterations — the structural design was right from commit 1; the iterations were all edge-handling (memory schema, FTS5 column rename, FTS5 query semantics). That's exactly the right shape for a benchmark-grade rewrite.

PR #236, opened 2026-05-17
bench(memory): add blind question-authoring protocol

usercise:
Refs #226

- Create bench/memory/README.md documenting the 5-step blindness protocol
- Create bench/memory/author_questions.py to export raw git log as source material
- Questions must be authored from raw commits, not from spelunk memory
- Ground truth = commit SHA, recorded with reviewer audit field
- Actual question sets (>=30 questions across >=3 repos) tracked separately

Post-review fixes:
- Check git log return code in author_questions.py
- Add reviewed_by field to question format
- Add TODO(#228) marker for Cross-Session Handoff section

### DIFF
--- a/bench/memory/README.md
+++ b/bench/memory/README.md
@@ -21,13 +21,15 @@ database. The protocol:
 3. **Ground truth:** For each question, record the commit SHA(s) that best
    answer it. Derive this from the raw git log, NOT from memory entries.
 4. **Review:** Have the question set reviewed by a second party with no
-   access to the spelunk memory database.
+   access to the spelunk memory database. Record the reviewer and date
+   in the `reviewed_by` field.
 5. **Format:** Save as `bench/memory/questions-<repo>.json`:
    ```json
    [
        {
            "question": "How does error handling work in the parser?",
-           "ground_truth_commit": "abc123def456"
+           "ground_truth_commit": "abc123def456",
+           "reviewed_by": "<name> on <date>"
        }
    ]
    ```
@@ -74,7 +76,7 @@ python bench/memory/decision_archaeology.py \
 
 ## Cross-Session Handoff
 
-See `bench/memory/cross_session_handoff.py`. Under redesign per issue #228.
+See `bench/memory/cross_session_handoff.py`. TODO(#228): update once redesign lands.
 
 ### Current limitations
 
--- a/bench/memory/README.md
+++ b/bench/memory/README.md
@@ -1,0 +1,82 @@
+# Memory Benchmarks
+
+## Decision Archaeology
+
+Measures whether spelunk memory can retrieve design rationale from git history
+better than lexical search (grep, FTS5).
+
+### Blindness Protocol
+
+Questions MUST be authored without access to the harvested spelunk memory
+database. The protocol:
+
+1. **Source material:** Read raw `git log` output for the target repo. Use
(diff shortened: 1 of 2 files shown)


usercise (commented):
**Status: scaffolding looks good, but the `Closes #226` keyword is premature**

The protocol doc and authoring tool are well-shaped — happy to see those land. But this PR doesn't satisfy #226's acceptance criteria, and merging it as-written will close the issue without the work being done.

## What's missing from #226

The issue asked for two things:

1. **A blindness protocol** for authoring questions independently of the harvested memory corpus. ✅ This PR delivers that — clear 5-step process in `bench/memory/README.md`, plus `author_questions.py` to export raw `git log` as the source material the author should read.
2. **The actual question sets:** *\"≥30 questions per repo … ≥3 repos with rich histories.\"* ❌ Not in this PR.

The only question set in the repo is still `bench/memory/questions-ripgrep.json` — 5 questions, all derived from spelunk memory entries (the exact circular setup #226 was filed to eliminate). Without new question sets authored under the blindness protocol, the four-baseline rerun in #235 has nothing rigorous to evaluate against.

## Suggested resolution

**Option A (recommended):** change the PR body from `Closes #226` to `Refs #226`, merge this as infrastructure, and open a follow-up issue *\"bench(memory): author blind question sets for ripgrep / tokio / ruff\"* tracking the actual data-authoring task. This unblocks others without misrepresenting #226's status.

**Option B:** hold this PR and author the question sets in the same PR. Probably 1-2 days of work depending on how much manual review goes into the questions.

I'd lean A — the protocol doc shouldn't block on what is effectively a content-creation deliverable.

## Nits

- `author_questions.py:127` doesn't check `result.returncode`. If `git log` fails (e.g. shallow clone, bad `--max-count` interaction) you'd silently get an empty `commits` list. A `subprocess.CalledProcessError` raise on non-zero, or at least a print-to-stderr, would help debugging.
- The protocol mentions \"have the question set reviewed by a second party with no access to the spelunk memory database\" — worth documenting *how* that review is recorded (e.g. a `reviewed_by` field in the questions JSON, or a checked-in review log) so the blindness claim is auditable later.
- Minor: the README says \"Under redesign per issue #228\" for Cross-Session Handoff. Once #228 lands, that line should update — easy to forget. Add a `TODO(#228)` marker maybe.

Otherwise, the scaffolding here is the right shape and I'd happily approve once the `Closes` → `Refs` swap is in.

usercise (commented):
**Status: looks good — ready to merge from my side. (Would be an `--approve` but GitHub doesn't allow approving a PR opened by the same user.)**

All review feedback addressed cleanly:

- ✅ `Closes #226` → `Refs #226` in the PR body — the issue stays open to track the actual blind question-set authoring.
- ✅ `reviewed_by` field added to the questions schema and documented in `bench/memory/README.md` step 4 + the example JSON block. Makes the blindness claim auditable.
- ✅ `author_questions.py` now raises on non-zero `git log` exit code instead of silently producing empty output.
- ✅ `TODO(#228)` marker on the Cross-Session Handoff section so the stale \"Under redesign\" line gets cleaned up when #228 lands.

Scaffolding is in the right shape and unblocks the next step: authoring the actual question sets under the documented protocol.

Note that #226 stays open after this merges — the issue's substantive deliverable (≥30 questions × ≥3 repos) is content authoring, not code, and should be done in a separate follow-up.

PR #238, opened 2026-05-17
bench(repobench): add --repo-filter for indexed-repo overlap

usercise:
Closes #225

- Add --repo-filter flag to sub-select RepoBench tasks to a single repo (Option B)
- Add indexed_repo_overlap_pct and repo_filter to output JSON
- Create bench/gemma/crosscodeeval/README.md documenting Options A/B/C
- Mark stale deepseek baseline in baselines/README.md

### DIFF
--- a/baselines/README.md
+++ b/baselines/README.md
@@ -11,6 +11,7 @@ spuriously invalidate itself.
 | File | Benchmark | Model | Status |
 |------|-----------|-------|--------|
 | `repobench-gemma-4-e2b-it-baseline.json` | RepoBench-Python | gemma-4-e2b-it | Superseded — regenerate with deepseek-v4-flash |
+| `repobench-deepseek-v4-flash-baseline.json` | RepoBench-Python | deepseek-v4-flash | **STALE** — pre-control run, 0% overlap. Delete after #225 rerun. |
 | `swebench-local-gemma-4-e2b-it-baseline.json` | SWE-bench local | gemma-4-e2b-it | Superseded — regenerate with deepseek-v4-flash |
 
 ## When to regenerate
--- a/bench/gemma/crosscodeeval/README.md
+++ b/bench/gemma/crosscodeeval/README.md
@@ -1,0 +1,53 @@
+# RepoBench Cross-File Completion
+
+Evaluates spelunk's semantic search against RepoBench-Python cross-file
+completion tasks.
+
+## Conditions
+
+| Condition | API calls | Tools | Purpose |
+|-----------|-----------|-------|---------|
+| `baseline_single_shot` | 1 | none | Single-shot completion, no loop |
+| `multi_turn_no_tools` | <=5 | none | Same loop budget, no retrieval |
+| `naive_search` | <=5 | `read_file`, `run_grep` | Loop with grep-level tools |
+| `spelunk` | <=5 | `spelunk_search` | Loop with semantic search |
+
+## Indexed repo mismatch
+
+The RepoBench dataset spans 1,751 repos; no single repo has more than 86 tasks.
+Spelunk indexes a single repo, so naive sampling produces near-0% overlap.
+
+### Option A: Per-task indexing (most accurate, heavy)
+
+Clone and index the source repo for each sampled task. ~50 clones +
+index runs. Many repos may have moved or disappeared.
+
+### Option B: Sub-select to one repo (recommended)
+
+Filter tasks to a single repo we have indexed.
+
+Recommended repo: `mpenning/ciscoconfparse2` (86 tasks in `cross_file_first`,
+the largest single-repo slice of the 8,033-task dataset). Next largest:
+`MarilynKeller/aitviewer-skel` (64), `Jisshubot/JISSHU_BOT` (63).
+For `--samples 50` runs, ccp2 is the only repo with comfortable headroom.
+
+```bash
+git clone https://github.com/mpenning/ciscoconfparse2.git /tmp/ccp2
+spelunk index /tmp/ccp2
+
+python bench/gemma/crosscodeeval/evaluate.py \
+    --condition spelunk \
+    --repo-path /tmp/ccp2 \
+    --repo-filter mpenning/ciscoconfparse2 \
+    --samples 50 \
+    --out bench/results/repobench-spelunk-ccp2.json
+```
+
+The `--repo-filter` flag ensures 100% overlap. The script cross-validates
+against the git remote of `--repo-path` and warns on mismatch. Output
+includes measured `indexed_repo_overlap_pct`.
+
+### Option C: Transfer test (lightest)
+
+Use any indexed repo without filtering. Frame results as a
+transfer/generalisation test. Overlap will be near 0%.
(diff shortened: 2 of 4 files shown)


usercise (commented):
**Status: structurally correct, but `indexed_repo_overlap_pct` is hardcoded rather than measured, which makes the field misleading.**

Good shape overall — `--repo-filter` is the right shape of mechanism (cheaper than per-task cloning, more honest than the previous setup), README clearly lays out A/B/C, and flagging the stale `repobench-deepseek-v4-flash-baseline.json` in `baselines/README.md` is exactly the right hygiene. Also verified independently that `mpenning/ciscoconfparse2` is the right Option B pick — it has 86 tasks in `cross_file_first`, the largest of any repo in the dataset (next is 64, then 63). So the README's recommendation is well-chosen.

## Concern 1 — `indexed_repo_overlap_pct` is intent, not measurement

`evaluate.py:547`:

```python
overlap_pct = 100.0 if repo_filter else 0.0
```

The field name implies a measured overlap percentage, but it's actually a boolean rephrased as 100/0 based on whether `--repo-filter` was passed. Two cases where this is silently wrong:

1. `--repo-filter mpenning/ciscoconfparse2 --repo-path /tmp/scikit-learn` (filter on, but indexed repo is unrelated). Reports 100% overlap; reality is 0%. The spelunk_search tool searches scikit-learn while tasks are from ccp2.
2. `--repo-path /tmp/ciscoconfparse2` *without* `--repo-filter` (user just runs it). Reports 0% overlap; reality is whatever fraction of sampled tasks happen to be from ccp2 (could be substantial since it's 86/8033 = ~1% of the dataset).

This becomes a problem later — when #232 (report rewrite) cites `indexed_repo_overlap_pct: 100` from the run, a reader who audits will find it's not derived from any measurement.

**Fix (recommended):** actually compute the overlap by reading the git remote of `--repo-path`:

```python
def detect_repo_owner_name(repo_path: Path) -> str | None:
    try:
        url = subprocess.run(
            ["git", "remote", "get-url", "origin"],
            cwd=repo_path, capture_output=True, text=True, timeout=5,
        ).stdout.strip()
        m = re.search(r"[:/]([^/:]+)/([^/]+?)(?:\.git)?$", url)
        return f"{m.group(1)}/{m.group(2)}" if m else None
    except Exception:
        return None

# in evaluate_split, after `sampled = dataset.select(indices)`:
indexed_name = detect_repo_owner_name(repo_path) if repo_path else None
hits = sum(1 for s in sampled if s["repo_name"] == indexed_name) if indexed_name else 0
overlap_pct = (hits / len(sampled)) * 100 if sampled else 0.0
```

Now the field reflects reality regardless of whether `--repo-filter` was used.

**Alternative (lighter):** rename the field to `repo_filter_applied: bool` so it reflects what's actually stored. Less informative, but honest about what it is.

## Concern 2 — no cross-validation between `--repo-filter` and `--repo-path`

Related to the above. If `--repo-filter X` and `--repo-path Y` point at different repos, the script silently runs anyway: dataset is filtered to X's tasks, but `spelunk_search` (and the `naive_search` tools) hit Y. The result is meaningless but the script doesn't say so.

If you add `detect_repo_owner_name` for Concern 1, this falls out naturally:

```python
if args.repo_filter and indexed_name and args.repo_filter != indexed_name:
    parser.error(
        f"--repo-filter ({args.repo_filter}) does not match the git remote "
        f"of --repo-path ({indexed_name}). The benchmark would search a "
        f"different repo than it samples from."
    )
```

A warning-only version (`print to stderr, continue`) is also reasonable if you want to leave room for advanced users running adversarial / transfer experiments.

## Nit — README should name the task counts

The README's Option B section says \"a well-represented repo\" but doesn't quantify. Worth adding a small fact:

> **Recommended repo:** `mpenning/ciscoconfparse2` is the largest single-repo slice of `cross_file_first` (86 tasks of 8,033). Next largest are `MarilynKeller/aitviewer-skel` (64) and `Jisshubot/JISSHU_BOT` (63). For `--samples 50` runs, ccp2 is the only repo with comfortable headroom.

Saves the next person from re-discovering this.

## Closing the issue

Reasonable to close #225 here — this is the mechanism + documentation half. The actual rerun (`spelunk index /tmp/ccp2 && evaluate --condition spelunk --repo-filter mpenning/ciscoconfparse2 --samples 50`) is implied by #232 (report rewrite). The stale-baseline flag in `baselines/README.md` is good prep for that.

Approving once Concern 1 lands (either the real measurement or the rename). Concern 2 and the README nit are non-blocking.

usercise (commented):
**Status: design is right, but `detect_repo_owner_name` is called twice and never defined — the script will `NameError` the moment anyone uses `--repo-filter` or `--repo-path`.**

The overlap-computation logic in `evaluate_split` (lines 548-553) and the cross-validation in `main` (lines 618-625) both call `detect_repo_owner_name(repo_path)`. That function isn't defined anywhere in `bench/gemma/crosscodeeval/evaluate.py`. AST check confirms — file defines 17 functions, none of them this one.

What happens in practice:

- `baseline_single_shot` / `multi_turn_no_tools` without `--repo-path` → still works (the `if repo_path else None` guards short-circuit), `overlap_pct = 0.0` in the JSON.
- `naive_search` or `spelunk` with `--repo-path` → crashes at the end of `evaluate_split` with `NameError: name 'detect_repo_owner_name' is not defined`.
- Any condition with `--repo-filter` and `--repo-path` → crashes at line 618 in `main` *before* the eval loop runs.

The headline scenario the PR was designed to enable (`--repo-path /tmp/ccp2 --repo-filter mpenning/ciscoconfparse2 --condition spelunk`) is the one that crashes hardest.

## Fix — add the function

The implementation from my earlier review just needs to land in the file. Tested it against your local ripgrep checkout:

```
>>> detect_repo_owner_name(Path('/Users/johan/opensource/ripgrep'))
'BurntSushi/ripgrep'
>>> detect_repo_owner_name(Path('/tmp'))
None
```

```python
def detect_repo_owner_name(repo_path: Path) -> str | None:
    """Detect 'owner/name' from the git remote origin URL of repo_path.

    Returns None if the repo has no origin remote, no git, or the URL
    doesn't match the github.com pattern (e.g. local-only clones).
    """
    try:
        url = subprocess.run(
            ["git", "remote", "get-url", "origin"],
            cwd=repo_path, capture_output=True, text=True, timeout=5,
        ).stdout.strip()
        m = re.search(r"[:/]([^/:]+)/([^/]+?)(?:\.git)?$", url)
        return f"{m.group(1)}/{m.group(2)}" if m else None
    except Exception:
        return None
```

Add it near `get_spelunk_version` / `scaffold_hash` (line ~450). `re` and `subprocess` are already imported.

## Smoke test before merge

A `python bench/gemma/crosscodeeval/evaluate.py --condition spelunk --repo-path /path/to/some/clone --samples 1 --repo-filter <whatever>` ought to run end-to-end and produce a result JSON. That's the test that would have caught this — easy to wire as a CI check once we have one.

## Otherwise

Everything else in the round looks good:

- Overlap calculation is now an actual measurement (`hits / len(sampled) * 100`) — exactly what concern 1 asked for.
- Cross-validation in `main` warns to stderr instead of erroring — that's the right call for transfer-test users running adversarially.
- Sanity that the eval still produces 0% overlap when `--repo-path` has no detectable origin remote.

Once the missing function lands and the smoke test passes, this is good to merge.

PR #242, opened 2026-05-17
bench(swebench): wire up Docker harness for resolve_rate

usercise:
Closes #229

- agent.py: add --save-patch flag to capture git diff after agent loop
- export_patches.py: collate batch results into SWE-bench prediction format
- swebench_eval.sh: orchestrate export + Docker harness evaluation

Follows the plan in tmp/swebench-evaluation-plan.md. The 26 missing repos need cloned before running; the harness wrapper accepts --dataset and --split for flexibility.

### DIFF
--- a/bench/agents/agent.py
+++ b/bench/agents/agent.py
@@ -516,12 +516,23 @@ def main() -> None:
     patch_path = None
     if args.save_patch:
         try:
-            diff = subprocess.run(
-                ["git", "diff"],
+            # Stage all changes (including untracked files from write_file)
+            # then diff against HEAD to capture everything.
+            subprocess.run(
+                ["git", "add", "-A"],
                 cwd=repo_path,
                 capture_output=True,
                 text=True,
                 timeout=30,
+                check=True,
+            )
+            diff = subprocess.run(
+                ["git", "diff", "--cached", "HEAD"],
+                cwd=repo_path,
+                capture_output=True,
+                text=True,
+                timeout=30,
+                check=True,
             ).stdout
             patch_path = Path(args.save_patch)
             patch_path.parent.mkdir(parents=True, exist_ok=True)
--- a/bench/agents/agent.py
+++ b/bench/agents/agent.py
@@ -473,6 +473,11 @@ def main() -> None:
     )
     parser.add_argument("--max-turns", type=int, default=20)
     parser.add_argument("--seed", type=int, default=42)
+    parser.add_argument(
+        "--save-patch",
+        default=None,
+        help="Save git diff to this file after agent finishes (for SWE-bench eval).",
+    )
     args = parser.parse_args()
 
     repo_path = Path(args.repo_path).resolve()
@@ -507,6 +512,23 @@ def main() -> None:
         seed=args.seed,
     )
 
+    # Save git diff for SWE-bench evaluation
+    patch_path = None
+    if args.save_patch:
+        try:
+            diff = subprocess.run(
+                ["git", "diff"],
+                cwd=repo_path,
+                capture_output=True,
+                text=True,
+                timeout=30,
+            ).stdout
+            patch_path = Path(args.save_patch)
+            patch_path.parent.mkdir(parents=True, exist_ok=True)
+            patch_path.write_text(diff)
+        except Exception as e:
+            print(f"Warning: failed to save patch: {e}", file=sys.stderr)
+
     # Reproducibility contract
     output = {
         "benchmark": "swebench-verified",
@@ -518,6 +540,7 @@ def main() -> None:
         "spelunk_version": get_spelunk_version(),
         "seed": args.seed,
         "max_turns": args.max_turns,
+        "patch_file": str(patch_path) if patch_path else None,
         **agent_result,
     }
     print(json.dumps(output))
--- a/bench/agents/batch_run.py
+++ b/bench/agents/batch_run.py
@@ -74,6 +74,8 @@ def run_task(
         str(max_turns),
         "--seed",
         str(seed),
+        "--save-patch",
+        str(SCRIPT_DIR.parent / "patches" / condition / f"{task_id}.patch"),
     ]
 
     try:
--- a/bench/agents/export_patches.py
+++ b/bench/agents/export_patches.py
@@ -1,9 +1,8 @@
 #!/usr/bin/env python3
 """Export agent patches to SWE-bench prediction format.
 
-Reads a batch result JSON (from batch_run.py or swebench_run.sh) and
-extracts per-task patches, model info, and reproducibility fields into
-the format expected by the SWE-bench Docker harness.
+Reads a batch result JSON and writes a flat predictions list the
+SWE-bench harness can consume, plus a metadata sidecar.
 
 Usage:
     python bench/agents/export_patches.py \\
@@ -36,18 +35,26 @@ def main():
     patches_dir = Path(args.patches_dir)
     predictions = {}
 
+    skipped = 0
+    errored = 0
+    no_patch = 0
+
     for r in results:
         task_id = r.get("task_id", "")
-        if r.get("skipped") or r.get("error"):
+        if r.get("skipped"):
+            skipped += 1
+            continue
+        if r.get("error"):
+            errored += 1
             continue
 
         patch_file = r.get("patch_file")
         if not patch_file:
+            no_patch += 1
             continue
 
         patch_path = Path(patch_file)
         if not patch_path.exists():
-            # Try relative to patches-dir
             patch_path = patches_dir / f"{task_id}.patch"
 
         if patch_path.exists():
@@ -56,22 +63,37 @@ def main():
                 "model_name_or_path": r.get("model", "deepseek-v4-flash"),
                 "model_patch": patch_path.read_text(),
             }
+        else:
+            no_patch += 1
 
-    output = {
-        "predictions": list(predictions.values()),
-        "metadata": {
-            "model": results[0].get("model", "") if results else "",
-            "condition": results[0].get("condition", "") if results else "",
-            "spelunk_version": results[0].get("spelunk_version", ""),
-            "seed": results[0].get("seed", ""),
-        },
-    }
+    predictions_list = list(predictions.values())
 
+    # Write flat list — SWE-bench harness expects this format
     Path(args.out).parent.mkdir(parents=True, exist_ok=True)
     with open(args.out, "w") as f:
-        json.dump(output, f, indent=2)
+        json.dump(predictions_list, f, indent=2)
 
-    print(f"Exported {len(predictions)} predictions to {args.out}")
+    # Metadata sidecar
+    meta_path = Path(args.out).with_suffix(".meta.json")
+    meta_path.write_text(
+        json.dumps(
+            {
+                "model": results[0].get("model", "") if results else "",
+                "condition": results[0].get("condition", "") if results else "",
+                "spelunk_version": results[0].get("spelunk_version", ""),
+                "seed": results[0].get("seed", ""),
+                "tasks_total": len(results),
+                "tasks_with_patch": len(predictions_list),
+                "tasks_skipped": skipped,
+                "tasks_errored": errored,
+                "tasks_no_patch": no_patch,
+            },
+            indent=2,
+        )
+    )
+
+    print(f"Exported {len(predictions_list)} predictions to {args.out}")
+    print(f"  Skipped: {skipped}  Errored: {errored}  No patch: {no_patch}")
 
 
 if __name__ == "__main__":
--- a/bench/agents/export_patches.py
+++ b/bench/agents/export_patches.py
@@ -1,0 +1,78 @@
+#!/usr/bin/env python3
+"""Export agent patches to SWE-bench prediction format.
+
+Reads a batch result JSON (from batch_run.py or swebench_run.sh) and
+extracts per-task patches, model info, and reproducibility fields into
+the format expected by the SWE-bench Docker harness.
+
+Usage:
+    python bench/agents/export_patches.py \\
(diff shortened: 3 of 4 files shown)


usercise (commented):
**Status: structure follows the plan, but three implementation gaps will stop it running end-to-end, and two of #229's acceptance criteria aren't yet met.**

The three new pieces line up with `tmp/swebench-evaluation-plan.md` step-for-step — agent saves a patch, exporter collates, wrapper invokes Docker. Good shape. The issues are in the wiring between them.

## Blocker 1 — predictions file is wrapped in metadata; SWE-bench harness won't parse it

`export_patches.py:114-122` writes:

```python
output = {
    "predictions": list(predictions.values()),
    "metadata": {...},
}
```

The official SWE-bench harness (`swebench.harness.run_evaluation`) expects `--predictions_path` to be either a flat JSON list `[{...}, {...}]`, a dict keyed by `instance_id`, or a JSONL file — *not* a wrapper object. Loading this file will either iterate over the dict keys (giving strings `"predictions"` and `"metadata"` instead of prediction objects) or fail outright.

Fix — write the list directly, and persist metadata as a sidecar:

```python
predictions_list = list(predictions.values())
Path(args.out).parent.mkdir(parents=True, exist_ok=True)
with open(args.out, "w") as f:
    json.dump(predictions_list, f, indent=2)

meta_path = Path(args.out).with_suffix(".meta.json")
meta_path.write_text(json.dumps({
    "model": results[0].get("model", "") if results else "",
    "condition": results[0].get("condition", "") if results else "",
    "spelunk_version": results[0].get("spelunk_version", ""),
    "seed": results[0].get("seed", ""),
    "tasks_total": len(results),
    "tasks_with_patch": len(predictions_list),
    "tasks_skipped": sum(1 for r in results if r.get("skipped")),
    "tasks_errored": sum(1 for r in results if r.get("error")),
    "tasks_no_patch": sum(1 for r in results if not r.get("skipped") and not r.get("error") and not r.get("patch_file")),
}, indent=2))
```

Worth verifying with a smoke test (see end) — if I'm wrong about the format I'd rather find out before this lands than after.

## Blocker 2 — `git diff` misses untracked files

`agent.py:521-524`:

```python
diff = subprocess.run(
    ["git", "diff"], cwd=repo_path, capture_output=True, text=True, timeout=30,
).stdout
```

`git diff` (no args) shows changes to **tracked** files only. Any new file the agent creates via the `write_file` tool — new test files, new utility modules, etc. — is silently dropped from the patch. For an agent with full `write_file` + `run_bash` access, this will absolutely happen on some tasks.

Cleanest fix is to stage everything and diff against `HEAD`:

```python
subprocess.run(["git", "add", "-A"], cwd=repo_path, check=True, timeout=30)
diff = subprocess.run(
    ["git", "diff", "--cached", "HEAD"],
    cwd=repo_path, capture_output=True, text=True, timeout=30,
).stdout
```

This mutates the index, but for SWE-bench-style scratch repos that's fine. If mutation is undesirable, the alternative is to enumerate untracked files via `git ls-files --others --exclude-standard` and append their contents as synthetic diff entries — much fiddlier.

Also worth checking returncode here — `subprocess.run` only raises if `check=True`; a failed `git diff` currently returns silent empty stdout.

## Blocker 3 — `batch_run.py` doesn't forward `--save-patch`

The wiring stops at `agent.py`. `bench/agents/batch_run.py` (the orchestrator used by `swebench_run.sh`) constructs the agent command without `--save-patch`, so a normal `python3 batch_run.py --condition spelunk_full ...` produces zero patch files. `export_patches.py` will then write an empty predictions list and the harness will have nothing to run.

Two reasonable shapes:

- **A:** `batch_run.py` auto-derives a per-task patch path (e.g. `bench/patches/<condition>/<task_id>.patch`) and unconditionally passes `--save-patch`. Simplest, most consistent.
- **B:** Add `--patches-dir` to `batch_run.py` and only pass `--save-patch` when set. More explicit, opt-in.

I'd lean A — patch files are cheap, having them always-on means the eval step doesn't depend on having predicted the eval at agent-run time.

## Issue #229 acceptance gaps

The issue called for two things this PR doesn't yet do:

1. **\"A harness wrapper that … merges resolve data back into the result JSON.\"** Right now `swebench_eval.sh` invokes the harness and leaves its output in `swebench_eval_outputs/spelunk-<ts>/`. The original `bench/results/swebench-<condition>-batch.json` still has `\"resolved\": false` for every task. Anyone reading the result JSON has to manually correlate it with the harness output dir. Worth adding a final step in the wrapper that reads the harness report and rewrites the result JSON with `resolved: <bool>` per task plus aggregate `resolve_rate` and `n_evaluated`.

2. **\"The 26/50 skipped tasks ('repo not found') are either resolved or explicitly excluded from the headline rate.\"** The PR body acknowledges \"the 26 missing repos need cloned before running\" but the wrapper doesn't surface this in the output. After eval, the merged JSON should make the denominator explicit:
   ```
   tasks_total: 50, tasks_evaluated: 24, tasks_skipped_pre_eval: 26,
   resolve_rate: <resolved>/<tasks_evaluated>
   ```
   Otherwise a reader will conflate \"resolve_rate = X% on SWE-bench Verified\" with \"X% on 24 of 50 tasks\" — exactly the kind of overstatement the original benchmark review flagged. The merged JSON also needs a `denominator_note` or similar making the 26-task gap explicit.

## Nits

- `run_id "spelunk-${TIMESTAMP}"` in `swebench_eval.sh` — loses condition info. With three conditions evaluated separately, you'll end up with three `spelunk-<ts>` runs and have to cross-reference timestamps to know which is which. Pull condition from the metadata sidecar (after Blocker 1) or accept `--condition` and use `spelunk-${CONDITION}-${TIMESTAMP}`.
- `export_patches.py` silently drops tasks with no `patch_file`. Print a short summary at the end: `\"Skipped N tasks (M skipped pre-eval, P errored, Q without patch)\"`. Without it, a 0-prediction file looks like a successful 0-prediction export rather than a misconfiguration.
- The plan suggested `pip install swebench`; worth confirming that's still the right package name (some forks renamed it `swe-bench` with a hyphen). A `command -v` check at the top of `swebench_eval.sh` would fail loudly if missing.

## Suggested smoke test before merge

Pick one task from the existing 24 OK results, rerun with `--save-patch /tmp/test.patch`, then:

```bash
bash bench/agents/swebench_eval.sh \\
    --results bench/results/swebench-baseline-20260515Tbatch.json \\
    --patches-dir /tmp/test-patches \\
    --max-workers 1
```

End-to-end run should complete and produce a per-instance pass/fail in the harness output. That single run would surface Blockers 1 and 2 immediately — Blocker 3 is detectable without even running (just grep `batch_run.py` for `--save-patch`).

Structural design is right; the fixes are all localised and small.

usercise (commented):
**Status: all three blockers fixed cleanly. One remaining concern around the resolve-merge step that needs a smoke test against the actual harness output before merging.**

The three blockers from the prior review are resolved:

- ✅ **Predictions file is now a flat list** with metadata moved to a `.meta.json` sidecar — exactly the shape `swebench.harness.run_evaluation` expects.
- ✅ **Patch capture uses `git add -A && git diff --cached HEAD`** with `check=True` on both subprocess calls — untracked files created by the agent's `write_file` are now included.
- ✅ **`batch_run.py` auto-derives a patch path per task** (`bench/patches/<condition>/<task_id>.patch`) and unconditionally passes `--save-patch`. Always-on, simplest shape.

Plus the nits got addressed: skipped/errored/no_patch counts are tracked and printed, run_id is now `spelunk-${CONDITION}-${TIMESTAMP}` derived from the metadata sidecar.

Good iteration.

## Remaining concern — the resolve-merge step

The new Step 3 in `swebench_eval.sh` is doing real work and addresses #229's \"merge resolve data back into the result JSON\" criterion. Two issues worth verifying before merge:

**Filename assumption may not match the harness.** Line 89-91:

```bash
HARNESS_DIR="swebench_eval_outputs/${RUN_ID}"
...
harness = json.load(open('${HARNESS_DIR}/results.json'))
```

The SWE-bench harness's output filename varies by version. Recent versions write per-instance reports under `<run_id>/<instance_id>/report.json` and a top-level summary that may be named `<model_name>.<run_id>.json` (at repo root, not inside `HARNESS_DIR`), or in older versions a `results.json` file at the run-dir root. I couldn't verify locally (swebench isn't installed here).

If the filename is wrong, the `if Path(...).exists() else {}` falls through silently — every task ends up with `resolved: False` and no error surfaces. Same class of silent-failure trap as the original FTS bug in PR #235.

**Two fixes — pick one:**

1. **Glob and inspect** — look for `*.json` under `HARNESS_DIR` and pick the file containing a `resolved` key. Robust against version drift.
2. **Assert + fail loudly** — `if [[ ! -f "$HARNESS_DIR/results.json" ]]; then echo "ERROR: harness output not found" >&2; exit 1; fi`. Surfaces the misnaming on first run rather than silently producing 0%-resolved JSON.

I'd lean (1) — version drift is real and the wrapper should be defensive.

**Aggregate fields aren't written to the result JSON.** The Python block prints `Evaluated: N  Resolved: M` to stdout, and adds per-task `resolved: <bool>` to each entry. But the headline aggregate — `tasks_total`, `tasks_evaluated`, `tasks_resolved`, `resolve_rate` — isn't persisted anywhere. A reader of `swebench-<condition>-batch.json` afterwards has to recompute the rate, and the 24/50 denominator (issue #229's #2 acceptance criterion) is invisible.

Suggest changing the result schema slightly. Either an envelope object:

```python
output = {
    "aggregate": {
        "tasks_total": len(results),
        "tasks_evaluated": tasks_evaluated,
        "tasks_resolved": tasks_resolved,
        "tasks_skipped_pre_eval": sum(1 for r in results if r.get("skipped")),
        "resolve_rate": tasks_resolved / tasks_evaluated if tasks_evaluated else 0,
    },
    "tasks": results,
}
json.dump(output, open('${RESULTS}', 'w'), indent=2)
```

Or a separate sidecar `<results>.resolve.json` so the per-task array format isn't disturbed.

Issue #229's second acceptance: \"The 26/50 skipped tasks ('repo not found') are either resolved or explicitly excluded from the headline rate.\" Currently the per-task `resolved` field is there but the explicit exclusion of the 26 skipped from the denominator isn't surfaced in any persisted output.

## Smoke test before merge

Pick one task that runs cleanly under baseline (any from the 24 OK ones), do an end-to-end:

```bash
# 1. Re-run one task with --save-patch via batch_run.py (now wired)
python3 bench/agents/batch_run.py --condition baseline --tasks 1 \
    --out /tmp/test-batch.json

# 2. Run the harness on it
bash bench/agents/swebench_eval.sh \
    --results /tmp/test-batch.json \
    --patches-dir bench/patches/baseline \
    --max-workers 1
```

End state should be: `swebench_eval_outputs/spelunk-baseline-<ts>/` populated, `/tmp/test-batch.json` updated with `resolved: true|false` per task. If the merge step silently no-ops, you'll see `resolved: false` and zero output from the `Evaluated: N  Resolved: M` print — that's the diagnostic.

Once the resolve-merge step is verified working end-to-end, this is ready.

usercise (commented):
**Status: ✅ both remaining concerns addressed cleanly. One small idempotency nit, otherwise ready to merge.**

The two prior concerns are fixed:

- ✅ **Harness output discovered by content, not by filename** — `find "$HARNESS_DIR" -maxdepth 2 -name '*.json' -exec grep -l '"resolved"' {} \;` is exactly the right shape. Version-drift resistant: as long as SWE-bench keeps emitting *some* JSON file with a `resolved` key, this will find it. Better than guessing the filename.
- ✅ **Aggregate fields persisted with explicit denominators** — the output envelope:

  ```json
  {
    "aggregate": {
      "tasks_total": N, "tasks_evaluated": N, "tasks_resolved": N,
      "tasks_skipped_pre_eval": N, "tasks_errored": N, "resolve_rate": 0.NNNN
    },
    "tasks": [...]
  }
  ```

  hits #229's explicit-denominator criterion exactly. A reader can't confuse \"X% on SWE-bench Verified\" with \"X% on 24/50 tasks\" because `tasks_evaluated` and `tasks_total` are right there.

- ✅ **Loud failure on missing harness output** — `echo "ERROR: no harness output with 'resolved' field under ${HARNESS_DIR}" >&2`. Silent no-op is gone.

## Small nit — wrap-twice idempotency

The script reads `${RESULTS}` (a flat list from `batch_run.py`), iterates over it, and writes back the envelope-wrapped shape `{aggregate: ..., tasks: [...]}`. If `swebench_eval.sh` is run a second time on the same `${RESULTS}` (e.g. to re-evaluate after a harness fix), the Python code does:

```python
results = json.load(open('${RESULTS}'))  # now a dict, not a list
for r in results:  # iterates over keys: "aggregate", "tasks"
    if not r.get('skipped'):  # AttributeError — r is a string
```

Confirmed:

```
$ python3 -c "import json; [print(f'iter: {r!r}') for r in json.load(open('/tmp/wrapped.json'))]"
iter: 'aggregate'
iter: 'tasks'
```

Won't crash on the `.get()` call (strings don't have `.get`, so it'd actually raise `AttributeError` and the bash `python3 -c ...` would exit non-zero, halting the wrapper). So the second run *would* fail loudly — but with a confusing error.

One-line fix at the top of the Python block:

```python
raw = json.load(open('${RESULTS}'))
results = raw['tasks'] if isinstance(raw, dict) and 'tasks' in raw else raw
```

Makes re-runs idempotent. Not blocking, but easy and saves a head-scratch later.

## Otherwise

Solid iteration. Three rounds was the right number — each one closed a real gap surfaced by the prior review's smoke test. After the idempotency tweak (or as a follow-up), this is good to ship.

The one remaining gate for #229's headline result is whoever clones the 26 missing repos so `tasks_evaluated` matches `tasks_total` of 50. That's content work, not code — could be its own follow-up issue, or folded into the existing rerun plan in #232.

PR #243, opened 2026-05-17
bench(memory): redesign Cross-Session Handoff with verifiable tasks

usercise:
Refs #228

Rewrite cross_session_handoff.py:
- Three Session 2 conditions: cold start, files present, with memory
- Session 1 forced cutoff with handoff prompt
- verify_cmd for binary pass/fail with output logging
- Pre-flight validation that verify_cmd fails on unmodified clone
- setup_cmd runs on all four clones
- Auto-cleanup for temp workdirs

Task corpus tracked in #247.

### DIFF
--- a/bench/memory/README.md
+++ b/bench/memory/README.md
@@ -76,9 +76,47 @@ python bench/memory/decision_archaeology.py \
 
 ## Cross-Session Handoff
 
-See `bench/memory/cross_session_handoff.py`. TODO(#228): update once redesign lands.
+Measures whether spelunk memory improves completion success for a fresh
+agent picking up partially-completed work.
 
-### Current limitations
+### Three conditions
 
-The 2026-05-15 proof-of-concept (n=1, no correctness check) is a demo,
-not a benchmark. See #228 for the redesign requirements.
+| Condition | Session 1 files | Memory access | Measures |
+|-----------|----------------|---------------|----------|
+| Cold start | None | None | Intrinsic task difficulty |
+| Files present | On disk | None | Value of file state alone |
+| With memory | On disk | Full | Value of files + memory |
+
+### Task format
+
+Tasks live in `bench/memory/handoff_tasks.json`:
+
+```json
+[
+    {
+        "task": "Fix the failing test in tests/test_parser.py",
+        "repo_url": "https://github.com/user/repo.git",
+        "setup_cmd": "pip install -e '.[dev]'",
+        "verify_cmd": "python -m pytest tests/test_parser.py -x -q"
+    }
+]
+```
+
+- `task`: natural-language description
+- `repo_url`: git clone URL (repo cloned fresh per task)
+- `setup_cmd`: shell command to install dependencies
+- `verify_cmd`: shell command that exits 0 on success (binary pass/fail)
+
+Session 1 is cut off at `--session-1-turns` (default 5) with a system
+prompt instructing the agent to write a detailed handoff. Three Session 2
+clones then attempt the task under the three conditions.
+
+Usage:
+
+```bash
+python bench/memory/cross_session_handoff.py \
+    --tasks bench/memory/handoff_tasks.json \
+    --session-1-turns 5 \
+    --session-2-turns 15 \
+    --out bench/results/handoff.json
+```
--- a/bench/memory/cross_session_handoff.py
+++ b/bench/memory/cross_session_handoff.py
@@ -1,39 +1,44 @@
 #!/usr/bin/env python3
-"""Cross-session handoff benchmark — measure the value of spelunk memory.
+"""Cross-session handoff benchmark — verifiable tasks, forced cutoff, n>=10.
 
-Simulates the core agent workflow:
-  1. Session 1: Agent works on a multi-file task, stores handoff + decisions
-     via `spelunk memory add` when it reaches turn limit.
-  2. Session 2: Fresh agent (no prior conversation) attempts the same task.
-     - Without memory: baseline tools only
-     - With memory: has spelunk_memory_search to find the handoff
+Three conditions for Session 2:
+    no_memory_no_session_1 — cold start, clean repo, no prior session
+    no_memory_present      — Session 1 file changes on disk, no memory tools
+    with_memory            — full spelunk memory access
 
-Measures: turns to completion, tokens used.
+Each task has a verify_cmd that produces a binary pass/fail signal.
+Tasks are defined in bench/memory/handoff_tasks.json.
 
 Usage:
     python bench/memory/cross_session_handoff.py \\
-        --repo-path /path/to/repo \\
-        --task "Rename all occurrences of foo to bar across the codebase" \\
+        --tasks bench/memory/handoff_tasks.json \\
         --model deepseek-v4-flash \\
-        --max-turns 15 \\
-        --out bench/results/handoff-demo.json
+        --session-1-turns 5 \\
+        --session-2-turns 15 \\
+        --out bench/results/handoff.json
 
-Prerequisites:
-    - Repo must be indexed: spelunk index <repo>
-    - API key in DEEPSEEK_API_KEY or passed via --api-key
+Workflow per task:
+  1. Clone fresh repo, run setup_cmd (installs deps)
+  2. Session 1: agent runs --session-1-turns, forced cutoff, stores handoff
+  3. Session 2a: fresh clone, no memory, no S1 files
+  4. Session 2b: S1 files on disk, no memory tools
+  5. Session 2c: S1 files + spelunk memory access
+  6. Verify each session's outcome with verify_cmd
 """
 
 import argparse
 import json
 import os
+import shutil
+import statistics
 import subprocess
 import sys
+import tempfile
 import time
 from pathlib import Path
 
 from dotenv import load_dotenv
 
-# Auto-load .env.local
 _root = Path(__file__).resolve().parents[2]
 _dotenv = _root / ".env.local"
 if _dotenv.exists():
@@ -41,6 +46,13 @@ if _dotenv.exists():
 
 AGENT_SCRIPT = Path(__file__).resolve().parents[1] / "agents" / "agent.py"
 REQUIREMENTS = Path(__file__).resolve().parents[1] / "requirements.txt"
+
+SESSION_1_SYSTEM_ADDON = (
+    "You have LIMITED TURNS. At the end of your turns, you MUST write a "
+    "detailed handoff summarising what you investigated, what you changed, "
+    "what is left to do, and any decisions you made. The next agent will "
+    "rely on this handoff to continue your work."
+)
 
 
 def run_agent(
@@ -51,12 +63,14 @@ def run_agent(
     api_base: str,
     api_key: str,
     max_turns: int,
-    extra_message: str = "",
+    seed: int,
+    extra_system: str = "",
 ) -> dict:
-    """Run agent.py and return the parsed JSON result."""
     issue_file = repo_path / "ISSUE.txt"
-    if not issue_file.exists():
-        issue_file.write_text(task)
+    task_text = task
+    if extra_system:
+        task_text = extra_system + "\n\n" + task
+    issue_file.write_text(task_text)
 
     cmd = [
         "uv",
@@ -83,31 +97,28 @@ def run_agent(
         "--max-turns",
         str(max_turns),
         "--seed",
-        "42",
+        str(seed),
     ]
-
     result = subprocess.run(cmd, capture_output=True, text=True, timeout=600)
     if result.returncode == 0:
         for line in reversed(result.stdout.strip().splitlines()):
             line = line.strip()
             if line.startswith("{"):
                 return json.loads(line)
-    return {
-        "error": True,
-        "stderr": result.stderr[:500],
-        "stdout": result.stdout[-500:],
-    }
+    return {"error": True, "stderr": result.stderr[:500]}
 
 
-def store_handoff(repo_path: Path, task: str, turns_used: int) -> None:
-    """Store a handoff memory entry via spelunk memory add."""
-    title = f"Handoff: {task[:80]}"
-    body = (
-        f"Task: {task}\n"
-        f"Session ran for {turns_used} turns before reaching limit.\n"
-        f"Repo state: partial changes may be present.\n"
-        f"Next agent should review current state and continue the work."
-    )
+def run_verify(repo_path: Path, cmd: str) -> bool:
+    try:
+        r = subprocess.run(
+            cmd, shell=True, cwd=repo_path, capture_output=True, text=True, timeout=60
+        )
+        return r.returncode == 0
+    except Exception:
+        return False
+
+
+def store_handoff(repo_path: Path, task: str, turns: int) -> None:
     subprocess.run(
         [
             "spelunk",
@@ -116,9 +127,14 @@ def store_handoff(repo_path: Path, task: str, turns_used: int) -> None:
             "--kind",
             "handoff",
             "--title",
-            title,
+            f"Handoff: {task[:80]}",
             "--body",
-            body,
+            (
+                f"Task: {task}\n"
+                f"Session cut off after {turns} turns.\n"
+                f"Repo state: partial changes present.\n"
+                f"Next agent: review current state and continue."
+            ),
         ],
         cwd=repo_path,
         capture_output=True,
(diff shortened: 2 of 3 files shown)


usercise (commented):
**Status: framework is exactly the right shape — but `Closes #228` is premature (placeholder tasks with fake URLs), and two correctness issues will silently produce noise.**

The redesign matches #228's spec well:

- ✅ Three Session 2 conditions: cold start / files present / with memory — matches the issue exactly.
- ✅ `--session-1-turns` forces a cutoff at a fixed turn count rather than letting the agent self-declare done.
- ✅ Session 1 system-prompt addon (`SESSION_1_SYSTEM_ADDON`) tells the agent it has limited turns and must write a detailed handoff. Good shape.
- ✅ `verify_cmd` produces binary pass/fail per task.
- ✅ Fresh clones per task in a `--workdir` (default `/tmp`). Hermetic.
- ✅ Aggregate success rates per condition printed at the end.

The three-condition shape in particular is what makes this a real benchmark instead of a demo — being able to attribute lift to *memory* vs *files on disk* is the differentiator.

## Premature `Closes #228` — placeholder task corpus

`handoff_tasks.json` has two tasks, both with:

```json
"repo_url": "https://github.com/user/repo.git"
```

That URL doesn't exist. Running this benchmark will produce a silent `git clone` failure for every task → 0% success across all conditions, looking like a memory-benchmark failure when it's actually no benchmark at all.

Issue #228's acceptance is:

> ≥10 task instances run across ≥2 repos, 3 conditions each.
> Per-task and aggregate success rates with confidence intervals.

This PR delivers the framework (the harder half — good) but not the corpus (the slower half — content authoring). Same pattern as PR #236 / #237 for #226, and PR #245 / pending-follow-up for #230.

**Recommended:** swap `Closes #228` → `Refs #228`, merge as scaffolding, and open a follow-up issue *\"bench(memory): author cross-session handoff task corpus (≥10 tasks across ≥2 repos with verify_cmd)\"*. Task authoring is real work — needs:

- Real repos with reproducible install paths.
- Tasks at the right difficulty: completable in 5 + 15 turns total but not trivial.
- `verify_cmd` that's deterministic and fast (≤60s — current `timeout=60` in `run_verify`).
- Pre-flight validation that `verify_cmd` returns non-zero in the unmodified state (else the success metric is degenerate).

## Correctness concern 1 — `setup_cmd` only runs on the template

`cross_session_handoff.py:337-340`:

```python
if setup_cmd != "true":
    subprocess.run(setup_cmd, shell=True, cwd=template, capture_output=True, timeout=120)
```

This runs setup *only* on `base/s1`. Then s2a is a separate fresh clone that **never has `setup_cmd` run on it**. For a task with `setup_cmd: \"pip install -e '.[dev]'\"`, s2a gets a clone without test dependencies installed. Its `verify_cmd` (e.g. `python -m pytest ...`) will fail with `ModuleNotFoundError` regardless of whether the agent solved the task — confounding the \"intrinsic task difficulty\" condition.

`s2b` and `s2c` get the s1 *files* via `copytree`, which includes installed `.egg-info` and `.venv` if those live in the project dir — but not if pip installed system-wide. Fragile.

Fix — run `setup_cmd` in all four clones:

```python
for clone_dir, label in [(base / "s1", "s1"), (base / "s2a", "s2a"), (base / "s2b", "s2b"), (base / "s2c", "s2c")]:
    ...  # clone
    if setup_cmd != "true":
        subprocess.run(setup_cmd, shell=True, cwd=clone_dir, capture_output=True, timeout=120)
```

Adds time but fixes the confound. Alternative: use a virtualenv per task and activate it inside `verify_cmd`, but that's a bigger restructure.

## Correctness concern 2 — `run_verify` discards output

`cross_session_handoff.py:185-193`:

```python
def run_verify(repo_path: Path, cmd: str) -> bool:
    try:
        r = subprocess.run(cmd, shell=True, cwd=repo_path, capture_output=True, text=True, timeout=60)
        return r.returncode == 0
    except Exception:
        return False
```

When `s2a_ok = False`, we have no idea why. Was it a syntax error in the task's `verify_cmd`? A flaky test? A genuine task failure? A missing dependency from concern #1?

For a benchmark where the outcome is a single bit per task, **the verify output is the diagnostic**. Save stdout+stderr (truncated) into the per-task result:

```python
def run_verify(repo_path: Path, cmd: str, log_path: Path | None = None) -> tuple[bool, str]:
    try:
        r = subprocess.run(cmd, shell=True, cwd=repo_path, capture_output=True, text=True, timeout=60)
        output = (r.stdout + r.stderr)[-2000:]
        if log_path:
            log_path.write_text(output)
        return r.returncode == 0, output
    except subprocess.TimeoutExpired:
        return False, "TIMEOUT after 60s"
    except Exception as e:
        return False, f"verify_cmd error: {e}"
```

And include the truncated `verify_output` (or `verify_log_path`) in each session's result dict. Without this, debugging task corpus issues becomes guesswork.

## Smaller concerns

- **No task validation pre-flight.** Before running any agent, validate that each task's `verify_cmd` returns non-zero on the unmodified clone. If it returns zero (task already solved) or errors out (broken verify_cmd), the metric is meaningless. One line: `assert not run_verify(fresh_clone, verify_cmd)[0], f\"Task {ti}: verify_cmd succeeds on unmodified repo\"`.

- **No 95% CI on success rates.** Issue #228 asked for \"per-task and aggregate success rates with confidence intervals.\" Currently only point estimates (e.g. `success=60% (6/10)`). For n=10, the Wilson interval on 6/10 is about [31%, 83%] — a wide enough band that publishing the point estimate without it would be the same overstatement risk that triggered the original review. Easy to add via `statsmodels.stats.proportion.proportion_confint` or by hand (Wilson is closed-form).

- **`workdir` is never cleaned up.** Default `tempfile.mkdtemp` produces `/tmp/handoff-XXXX/` per run, each holding 4 git clones (× number of tasks). After a few runs this gets large. Add `atexit.register(shutil.rmtree, workdir, ignore_errors=True)` when `--workdir` isn't explicitly passed, and leave alone when the user provided one.

- **No `--workers` for parallelism.** 10 tasks × 4 agent runs × ~2 min ≈ 80 min serial. Could parallelise across tasks (different workdirs are hermetic). Not a blocker — task corpus is the gate, not throughput. Worth a follow-up issue.

- **`uv run --with-requirements` invoked per agent run.** Each subprocess sets up env from scratch. Slow but predictable. Acceptable.

## Suggested smoke test before merge

Even with placeholder URLs, the framework can be exercised by writing one real task (e.g. against `BurntSushi/ripgrep` with a simple `verify_cmd: \"cargo check\"`). End-to-end run with `--workdir /tmp/handoff-test`:

```bash
python bench/memory/cross_session_handoff.py \\
    --tasks /tmp/one_task.json --workdir /tmp/handoff-test \\
    --session-1-turns 3 --session-2-turns 5 \\
    --out /tmp/handoff-test.json
```

End state should have:
- 4 clones under `/tmp/handoff-test/task0/{s1,s2a,s2b,s2c}/`
- A handoff stored visible via `cd /tmp/handoff-test/task0/s1 && spelunk memory list`
- `/tmp/handoff-test.json` with 3 condition success rates

That smoke test would surface concern #1 (setup) and #2 (no debug output) immediately.

## What I'd like to see before final approval

Either:

- **A:** swap `Closes` → `Refs`, address concerns 1 & 2 (setup_cmd + verify logging), file the corpus-authoring follow-up issue. Smaller delta, scaffolding-merge model that #226/#236 used.
- **B:** also author 2-3 real tasks and replace the placeholders before merge. Bigger delta, more confidence.

I'd lean A given the pattern with prior PRs — keep the scaffolding-merging cadence consistent. The follow-up issue then becomes the real-work tracker.

usercise (commented):
**Status: ✅ all blocking concerns from the prior round addressed. Two smaller issues worth knowing about — one methodological (subtle but real), one a missing-feature follow-up.**

Prior round feedback all landed:

- ✅ **`Closes #228` → `Refs #228`** with [#247](https://github.com/usercise/spelunk/issues/247) tracking the task corpus authoring. Same scaffolding-merge model as #237 / #248. Verified #247 is filed.
- ✅ **`setup_cmd` runs on all four clones** — `for clone_dir in clones.values(): ... setup_cmd runs on each`. The S2a/S2b/S2c confounds from missing deps are gone.
- ✅ **`run_verify` returns `(bool, str)`** with truncated output; each session result now stores `verify_output` in the JSON, and the console prints the first 60 chars. Plus proper `TimeoutExpired` vs generic `Exception` handling. Debugging task failures is no longer blind.
- ✅ **Pre-flight validation** — `run_verify(clones["s2a"], verify_cmd)` runs *before* any agent. Warns on degenerate tasks (verify passes unmodified), confirms expected failure otherwise.
- ✅ **`atexit.register(shutil.rmtree, ...)`** when `--workdir` isn't explicit. Auto-cleanup of temp dirs.

Good iteration.

## Methodological concern — S2b's \"no memory\" condition is leaky

`shutil.copytree(str(clones["s1"]), str(clones["s2b"]), symlinks=True)` copies *everything* from S1, including `.git/notes/` (where `spelunk memory add --kind handoff` persists the entries) and `.spelunk/` (the index DB).

The S2b condition gives the agent `baseline` tools (`read_file`, `run_bash`, `write_file`) without `spelunk_memory_search`. **But the agent has `run_bash`.** An agent that knows about spelunk could `run_bash spelunk memory search "..."` and bypass the no-memory condition entirely. The memory data is physically on disk under S2b's clone; only the dedicated *tool* is missing from the agent's tool list.

Whether this matters in practice depends on model behaviour — deepseek-v4-flash may or may not spontaneously try shelling out to spelunk. But the claim S2b tests (\"value of file state alone, without memory\") isn't cleanly isolated as written.

Two cheap fixes:

1. **Strip the memory + index after copy:**
   ```python
   shutil.copytree(str(clones["s1"]), str(clones["s2b"]), symlinks=True)
   shutil.rmtree(clones["s2b"] / ".spelunk", ignore_errors=True)
   subprocess.run(["git", "notes", "remove", "--ignore-missing", "HEAD"], cwd=clones["s2b"], capture_output=True)
   # Or more thorough: rm -rf .git/refs/notes/spelunk
   shutil.rmtree(clones["s2b"] / ".git" / "refs" / "notes", ignore_errors=True)
   ```

2. **Document the leak** in the README and accept it — argue that the agent doesn't typically discover the memory store on its own, and note this as a known limitation. Honest but not airtight.

I'd lean (1). The whole point of the three-condition structure is clean attribution, and a leaky S2b muddies that.

## Missing feature — 95% CI on success rates

[#228's](https://github.com/usercise/spelunk/issues/228) acceptance: *\"per-task and aggregate success rates with confidence intervals.\"*

Current output reports point estimates only:

```
  Cold start         success=20% (2/10)
  Files present      success=50% (5/10)
  With memory        success=70% (7/10)
```

At n=10, Wilson 95% CIs on those are roughly [3%, 56%], [24%, 76%], [40%, 89%] — wide enough that point-estimate-only reporting reproduces the same overstatement risk that prompted the original review.

Closed-form Wilson is ~5 lines (no scipy needed):

```python
def wilson_ci(successes: int, n: int, z: float = 1.96) -> tuple[float, float]:
    if n == 0:
        return (0.0, 0.0)
    p = successes / n
    denom = 1 + z**2 / n
    centre = (p + z**2 / (2 * n)) / denom
    margin = (z * (p * (1 - p) / n + z**2 / (4 * n**2)) ** 0.5) / denom
    return (max(0, centre - margin), min(1, centre + margin))
```

Print as `success=70% [40%, 89%] (7/10)` and include `success_ci_95: [low, high]` in the aggregate JSON. Non-blocking but worth before any number from this benchmark gets published.

## Smaller suggestions

- **Multi-trial per condition.** [#228](https://github.com/usercise/spelunk/issues/228) also asked for *\"ideally with 2–3 independent Session 2 runs per task (different temperature / seed).\"* Currently single-trial. A `--s2-trials N` flag with seeds `args.seed + trial_idx` would address this. Could fold into [#247](https://github.com/usercise/spelunk/issues/247) as a measurement-design follow-up rather than blocking this PR.
- **Cleanup-and-rerun race.** `if (clone_dir / ".git").exists(): skip clone, otherwise clone`. If a previous run was interrupted mid-clone, the partial clone is kept and never re-validated. Add a `git -C "$clone_dir" rev-parse HEAD >/dev/null 2>&1` sanity check, or just `shutil.rmtree(clone_dir); clone fresh`.

## What's good

- The dict-based `clones` refactor (`clones["s1"]`, etc.) is cleaner than the prior list-of-tuples.
- Pre-flight \"verify_cmd fails as expected\" log line is genuinely useful — gives confidence the test corpus is well-formed before burning API tokens.
- The verify_output truncation at `-2000:` (last 2k chars) is a sensible budget — captures the tail of pytest tracebacks where the actionable info usually lives.

After the S2b leak is plugged (or explicitly accepted with a docs note) and Wilson CIs are added, this is publish-ready.

usercise (commented):
**Status: Wilson CI lands cleanly; memory-strip is partial — `.spelunk/` removal alone doesn't fully isolate S2b because spelunk has a global memory fallback. Worth one more iteration or an explicit accept-and-document.**

## ✅ Wilson CI

Correct implementation of the closed-form Wilson interval, exactly the shape I suggested. Verified the math:

```python
def wilson_ci(successes: int, n: int, z: float = 1.96) -> tuple[float, float]:
    ...
```

Stored under `aggregate.<cond>_ci_95` in the JSON, printed inline next to the point estimate (`success=50% [24%, 76%] (5/10)`). Renaming the condition keys to snake_case (`cold_start`, `files_present`, `with_memory`) is a nice consistency cleanup. Good.

## ⚠️ Memory-strip is incomplete

The fix removes `.spelunk/` (per-project memory) and `git notes` (experimental backend) from the S2b clone — both correct. But spelunk has a **global memory store** at `~/.config/spelunk/memory.db` that's used as a fallback when no per-project store is reachable. Verified locally:

```
$ ls ~/.config/spelunk/
config.toml  index.db  memory.db (3.6 MB)  registry.db

$ cp -r ~/opensource/ripgrep /tmp/test-strip && rm -rf /tmp/test-strip/.spelunk
$ cd /tmp/test-strip && spelunk memory list --limit 1
[returns one entry — but from the GLOBAL store, not ripgrep's]
```

The returned entry was an unrelated `antipattern` memory from a different project. The agent in S2b has `run_bash`, so any `run_bash spelunk memory search "..."` query will hit the global store and may surface noise (or, worse, semantically-relevant entries from other projects the dev has worked on).

For the *strict* three-condition attribution test #228 was meant to enable, this means S2b is \"approximately memory-free\" rather than cleanly so. A reviewer arguing the benchmark is rigorous will spot this.

**Verified the override path:**

```
$ env HOME=/tmp/empty-home sh -c 'mkdir -p /tmp/empty-home/.config/spelunk && cd /tmp && spelunk memory list'
No memory entries found.
```

So fully isolating S2b requires either:

**Option A (recommended) — env-override in subprocess.** Pre-create an empty config dir, point the s2b agent's `HOME` at it:

```python
# Once, at the start of each task
s2b_empty_home = workdir / f"task{ti}" / "empty-home"
(s2b_empty_home / ".config" / "spelunk").mkdir(parents=True, exist_ok=True)

# In run_agent for s2b only — needs a `env` kwarg added to run_agent
env = os.environ.copy()
env["HOME"] = str(s2b_empty_home)
result = subprocess.run(cmd, env=env, capture_output=True, text=True, timeout=600)
```

Adds one parameter to `run_agent`, ~5 lines total.

**Option B — temporarily move `~/.config/spelunk/memory.db` aside** during the s2b run. Brittle (races if multiple benchmark runs go in parallel), but no code changes to `run_agent`.

**Option C — accept and document.** Add to `bench/memory/README.md` and to a `Limitations` block in the result JSON: \"S2b strips per-project memory but spelunk's global memory store at `~/.config/spelunk/memory.db` remains reachable via the agent's `run_bash` tool. In practice the global store contains entries from unrelated projects which the agent rarely incorporates, but the no-memory condition is not strictly isolated.\"

I'd lean A — the whole point of the three-condition structure is clean attribution, and a 5-line subprocess env override is cheaper than carrying a footnote forever. But C is a reasonable pragmatic choice if you'd rather ship now and tighten later.

## Smaller nit

The git-notes strip is defensive (catches the experimental `--backend git-notes` if someone uses it) but the comment doesn't say so. A one-line `# Belt-and-suspenders: also clear git notes in case --backend git-notes is in use.` would prevent a future maintainer wondering why we strip both.

## Summary

| Concern | Status |
|---|---|
| Wilson CI | ✅ added and correct |
| S2b memory leak | ⚠️ partial — global store still reachable via `run_bash` |
| Workdir cleanup | ✅ (prior round) |
| Pre-flight validation | ✅ (prior round) |
| `Closes` → `Refs` + #247 | ✅ (prior round) |

After the env-override (or the docs note), this is ready. The benchmark is now actually fit for purpose, modulo the leak. Solid landing across three rounds.

PR #244, opened 2026-05-17
bench(perf): add perf_scale.sh orchestrator for multi-repo scale benchmarks

usercise:
Refs #231

Create bench/perf_scale.sh orchestrator:
- Labelled repo sizes via --repos
- Indexing timed with throughput (files/sec, chunks/sec)
- Search latency across 2/7/16 word queries
- Optional memory benchmark
- Host/CPU info for reproducibility
- Aggregate JSON output

Scale execution tracked in follow-up issue.

### DIFF
--- a/bench/perf_scale.sh
+++ b/bench/perf_scale.sh
@@ -93,8 +93,8 @@ for entry in "${REPO_ENTRIES[@]}"; do
 
     # Count files and chunks via spelunk status
     STATS=$(cd "$REPO_DIR" && "$SPELUNK" status --format json 2>/dev/null || echo '{"files":0,"chunks":0}')
-    FILES=$(echo "$STATS" | python3 -c "import json,sys; print(json.load(sys.stdin).get('files',0))")
-    CHUNKS=$(echo "$STATS" | python3 -c "import json,sys; print(json.load(sys.stdin).get('chunks',0))")
+    FILES=$(echo "$STATS" | python3 -c "import json,sys; print(json.load(sys.stdin).get('file_count',0))")
+    CHUNKS=$(echo "$STATS" | python3 -c "import json,sys; print(json.load(sys.stdin).get('chunk_count',0))")
     echo "    ${FILES} files, ${CHUNKS} chunks"
 
     # Search latency — 2/7/16 word queries
--- a/bench/perf_scale.sh
+++ b/bench/perf_scale.sh
@@ -1,8 +1,8 @@
 #!/usr/bin/env bash
 # bench/perf_scale.sh — orchestrator for scale-level performance benchmarks
 #
-# Runs indexing, search, and memory benchmarks across multiple repo sizes
-# and aggregates results into a single JSON file.
+# Runs indexing timing, search latency, and optional memory benchmarks
+# across multiple labelled repo sizes, aggregating into a single JSON.
 #
 # Usage:
 #   bash bench/perf_scale.sh --repos small:path medium:path large:path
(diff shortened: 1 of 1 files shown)


usercise (commented):
**Status: cannot ship as written — three real bugs, one of which prevents the script from running at all, plus one of issue #231's primary acceptance criteria is unmet.**

The shape is right (orchestrator across labelled repo sizes, aggregate JSON output, optional memory benchmark, skip-flag). Each individual problem is small.

## Blocker 1 — `PATH` variable shadowing breaks the script

`perf_scale.sh:72`:

```bash
PATH="${entry#*:}"
```

This **overrides the shell's `$PATH` environment variable** with the repo's filesystem path (e.g. `/Users/johan/opensource/ripgrep`). After this line, every unqualified command in the loop — `find`, `python3`, `wc`, `tr`, `bash` — fails with `command not found`, because `find` is no longer at `/<repo-path>/find`. Confirmed by reproducing:

```
$ bash -c 'PATH="/tmp"; find /tmp -type d | head -2'
bash: head: command not found
```

Fix: rename to anything else, `REPO_PATH`, `REPO_DIR`, etc. Everywhere it's used as a directory argument (lines 74, 75, 79, 86-87, 103) needs the rename too.

## Blocker 2 — `spelunk chunks` invoked on a directory; it wants a file

`perf_scale.sh:87`:

```bash
CHUNKS=$("$SPELUNK" chunks --format json "$PATH" 2>/dev/null | python3 ... || echo "?")
```

Per `spelunk chunks --help`: `Arguments: <PATH>  File path (exact or suffix match against indexed paths)`. It produces chunks for a single file. Passing a directory will error, the `|| echo "?"` swallows it, and `chunks` ends up as the string `"?"` in the output JSON.

Use `spelunk status` instead — it reports file and chunk counts for the current project's index. JSON-mode output exists (`--format json`). Run from inside the repo:

```bash
STATS=$(cd "$REPO_DIR" && "$SPELUNK" status --format json 2>/dev/null)
FILES=$(echo "$STATS" | python3 -c "import json,sys; print(json.load(sys.stdin).get('files', 0))")
CHUNKS=$(echo "$STATS" | python3 -c "import json,sys; print(json.load(sys.stdin).get('chunks', 0))")
```

Also: the current `find -type f -not -path '*/.git/*'` count doesn't match what spelunk actually indexes (spelunk respects `.gitignore`, `.spelunkignore`, etc.). Using `spelunk status` makes the `files` field consistent with what the index actually contains.

## Blocker 3 — indexing isn't timed; #231's first acceptance criterion is unmet

`perf_scale.sh:82-83`:

```bash
echo "  Indexing..."
"$SPELUNK" index "$PATH" 2>&1 | tail -1 || true
```

The `spelunk index` call is invoked but the duration is thrown away. Issue #231's first acceptance criterion is **\"Indexing numbers for ≥3 repo sizes including one ≥10k files\"** — that's an indexing-throughput benchmark, not a search benchmark. As written, the script only reports search latency. The indexing-at-scale signal is the most important data point for the perf rewrite (since incremental indexing is one of spelunk's product-defining properties).

Fix — wrap the index call in `time`:

```bash
INDEX_START=$(python3 -c "import time; print(time.monotonic())")
"$SPELUNK" index "$REPO_DIR" >/dev/null 2>&1 || true
INDEX_ELAPSED=$(python3 -c "import time; print(round(time.monotonic() - $INDEX_START, 2))")
```

And record `index_seconds` in the per-repo block alongside `files`, `chunks`, `search`. Files-per-second and chunks-per-second can be computed from those at report time.

## Smaller concerns

- **Query length matrix.** The three hard-coded queries (`error handling`, `parse command line`, `validate configuration input`) are 2 / 3 / 3 words — not three distinct lengths. The previous `perf_search.sh` table in the report used 2 / 7 / 16 words. Match that for continuity, or document why this differs.
- **Reproducibility fields missing.** No `host` (Apple M-series vs Linux box matters a lot for perf numbers), no `seed`, no `scaffold_hash`. Perf numbers without machine context are barely reproducible. At minimum capture `uname -a` and CPU info via `sysctl -n machdep.cpu.brand_string` (macOS) / `/proc/cpuinfo` (Linux). The Limitations section in PR #246's report does call out single-machine perf as a known issue, but that doesn't excuse the result file not recording it.
- **`spelunk_version` capture is brittle.** Line 66 does `$("$SPELUNK" --version 2>&1 | head -1)` — if spelunk isn't installed, the captured string becomes `command not found` and gets embedded into JSON, producing invalid JSON. Add a `command -v spelunk` check at the top of the script.
- **`set -euo pipefail` with `|| true`.** The mixture is fine but means failed `spelunk index` invocations are silently treated as success. With Blocker 3's timing in place, a failed index produces `index_seconds` of a few milliseconds — easy to spot as bogus. Add `if ! "$SPELUNK" index ... ; then echo "FAILED" >&2; continue; fi`.

## Smoke test before merge

```bash
bash bench/perf_scale.sh --repos small:/Users/johan/opensource/ripgrep --skip-memory
```

With Blockers 1+2 fixed, this should produce a JSON file with non-zero `files`, non-`"?"` `chunks`, and three search latency entries. Today it'll fail with `command not found: find` at line 86.

Structural design is right; once these fixes land, this is a 5-minute follow-up.

usercise (commented):
**Status: 3 prior blockers all fixed, but one new silent-zero bug surfaced on smoke-test — and `Closes #231` is premature.**

The previous round's blockers are resolved cleanly:

- ✅ **`PATH` shadowing** — renamed to `REPO_DIR` everywhere.
- ✅ **`spelunk chunks <dir>` swapped for `spelunk status --format json`** — right command for the job.
- ✅ **Indexing is now timed** — `INDEX_START` / `INDEX_ELAPSED` recorded; `files_per_second` and `chunks_per_second` derived. Plus `set -e` style early-exit if the index fails.
- ✅ **Host/CPU captured** via `uname -a` + `sysctl/cpuinfo`.
- ✅ **`command -v spelunk` check** at script start.
- ✅ **Query length matrix** moved to 2 / 7 / 16 words, matching the existing `perf_search.sh` numbers in the prior report.

Good iteration — addresses every concern from the prior review.

## New blocker — field name mismatch on `spelunk status` output

Smoke-tested the new script:

```
$ bash bench/perf_scale.sh --repos small:/Users/johan/opensource/ripgrep --skip-memory --search-iters 2
...
  small: 0 files, 0 chunks, index=1.61s
```

But ripgrep is actually 122 files / 3,915 chunks. The script silently reports zero because the output schema doesn't match what the script reads:

```
$ cd /Users/johan/opensource/ripgrep && spelunk status --format json | head -5
{
  "chunk_count": 3915,
  "drift_candidates": [],
  "embedding_count": 3915,
  "file_count": 122,
```

`spelunk status --format json` emits `file_count` and `chunk_count`. The two Python one-liners at lines 73-74 read `files` and `chunks` instead, so both extract their default `0`. The resulting JSON has `files: 0, chunks: 0, files_per_second: 0, chunks_per_second: 0` for *every* repo — same class of silent-zero failure as PR #235's FTS bug, just with no exception path to log.

Fix:

```bash
FILES=$(echo "$STATS" | python3 -c "import json,sys; print(json.load(sys.stdin).get('file_count',0))")
CHUNKS=$(echo "$STATS" | python3 -c "import json,sys; print(json.load(sys.stdin).get('chunk_count',0))")
```

After the rename, the smoke test should report `122 files, 3915 chunks, ~75 files/sec, ~2400 chunks/sec` for ripgrep.

This is exactly the smoke test I suggested in the prior review (`bash bench/perf_scale.sh --repos small:ripgrep --skip-memory`). One more iteration of \"run it once before merging\" catches this.

## `Closes #231` is premature

Issue #231's acceptance:

> - Indexing numbers for ≥3 repo sizes including one ≥10k files.
> - Search latency on a repo with ≥10k chunks.
> - Memory benchmark actually run, not just \"infrastructure ready\".
> - All three scale runs reproducible from a single `bench/perf_scale.sh` orchestrator script.

The PR delivers the *orchestrator* (criterion 4) — good. But criteria 1, 2, 3 require **actually running the benchmarks at scale**, which this PR doesn't. There are no committed result JSONs in `baselines/` for small/medium/large repos.

Same pattern as #226/#236 → #237, and #230/#245 → #248. Recommend:

- Swap `Closes #231` → `Refs #231`.
- Open a follow-up issue *\"bench(perf): execute scale runs against small/medium/large repos and commit baselines\"*.
- That follow-up tracks: pick the three repos (the bench/agents/tasks_50.json checkouts under `~/opensource/spelunk-bench/repos/` could supply medium-size candidates — django at ~3.5k files), run the orchestrator, commit baselines.

## Otherwise

The host/cpu capture is a nice touch — once we have multi-machine perf data this will matter for cross-run comparison. Query length matrix matches the original `perf_search.sh` shape, so future-report side-by-side comparison is feasible.

After the field-name fix lands and the body is updated to `Refs`, this is good. One more re-run of the smoke test should confirm non-zero file/chunk counts before merging.

usercise (commented):
**Status: ✅ both prior issues addressed. Ready to merge from my side. One observation worth noting for the eventual scale-run follow-up.**

Both concerns from the prior round are fixed:

- ✅ **`file_count` / `chunk_count`** — field names corrected in both Python one-liners. Smoke-tested against ripgrep:
  ```
  small: 123 files, 3916 chunks, index=0.1s
    search (2 words) p50=412.8ms  p95=412.8ms
    search (7 words) p50=276.9ms  p95=276.9ms
    search (16 words) p50=319.1ms  p95=319.1ms
  ```
  Counts now match `spelunk status` output (was `0 files, 0 chunks` previously).

- ✅ **`Closes #231` → `Refs #231`** — body updated, #231 stays open as the umbrella for the actual scale runs.

## One observation — incremental indexing skews the throughput field

The smoke test above shows `index_seconds: 0.1`, `files_per_second: 1230.0`, `chunks_per_second: 39160.0`. That's because ripgrep was already indexed locally — spelunk's incremental logic skipped unchanged files. **Cold-start indexing of the same repo takes ~30-60 seconds** (per the original `perf_index.sh` numbers in the old report, where ripgrep indexed at ~3.9 files/sec, not 1230).

The result JSON doesn't distinguish cold-start from warm/incremental. Two consumers will read radically different numbers depending on whether the repo was previously indexed:

| Scenario | files_per_second |
|---|---|
| Fresh index (first-time) | ~3-5 |
| Incremental (warm) | ~1000+ |

Neither is wrong, but they measure different things. For the scale-run follow-up that #231 still tracks, suggest:

1. **Force cold-start by default:** `rm -rf "$REPO_DIR/.spelunk"` before `spelunk index`. Or add a `--force-reindex` flag.
2. **Record both:** time cold-start and an immediate re-index, capture both as `index_seconds_cold` / `index_seconds_incremental`. The incremental number is genuinely interesting for the \"agent re-runs after small code changes\" workflow.
3. **At minimum, record an `index_mode: \"cold\"|\"incremental\"` field** in the JSON so a reader can tell which they're looking at.

Not blocking — the orchestrator works correctly and the throughput math is right given its inputs. But before the scale runs go in `baselines/`, this distinction matters.

## Otherwise

Solid landing. Three rounds was the right number — each one closed a real, smoke-test-surfaced gap (PATH shadow → `chunks` invocation → field-name mismatch). The pattern of \"run it once before claiming done\" caught each in turn.

After this merges, #231 stays open for the actual scale execution (pick three repos, do cold-start indexing, commit baselines). That's a clean handoff to whoever picks up the rerun work alongside the corpus authoring tasks (#237, #247, #248).

PR #245, opened 2026-05-17
bench(graph): add code-graph benchmark — grep vs search vs graph

usercise:
Refs #230

Create bench/graph/evaluate.py with three conditions:
- grep: git grep with word boundaries, capped at k
- spelunk_search: semantic search
- spelunk_graph: code graph traversal (with stderr warnings on unexpected shapes)

Metrics: precision@k, recall@k, F1. No LLM required.
Task corpus tracked in #248.

### DIFF
--- a/bench/graph/evaluate.py
+++ b/bench/graph/evaluate.py
@@ -30,25 +30,28 @@ Task format (JSON):
 
 import argparse
 import json
+import re
 import statistics
 import subprocess
+import sys
 import time
 from datetime import datetime, timezone
 from pathlib import Path
 
 
-def run_grep(repo_path: Path, symbol: str) -> set[str]:
-    """Return set of file paths containing the symbol via git grep."""
+def run_grep(repo_path: Path, symbol: str, limit: int = 10) -> set[str]:
+    """Return set of file paths containing the symbol via git grep, capped at limit."""
     try:
         result = subprocess.run(
-            ["git", "grep", "-l", symbol],
+            ["git", "grep", "-l", "-E", f"\\b{re.escape(symbol)}\\b"],
             cwd=repo_path,
             capture_output=True,
             text=True,
             timeout=30,
         )
         if result.returncode == 0:
-            return set(result.stdout.strip().split("\n"))
+            files = result.stdout.strip().split("\n")
+            return set(files[:limit])
         return set()
     except Exception:
         return set()
@@ -84,15 +87,24 @@ def run_spelunk_graph(repo_path: Path, symbol: str, limit: int = 10) -> set[str]
         )
         if result.returncode == 0 and result.stdout.strip():
             results = json.loads(result.stdout)
-            files = set()
+            files_ordered = []
+            seen = set()
             for r in results if isinstance(results, list) else [results]:
+                if "edges" not in r and not isinstance(r, dict):
+                    print(
+                        f"  graph: unexpected shape for {symbol}: {str(r)[:120]}",
+                        file=sys.stderr,
+                    )
+                    continue
                 for edge in r.get("edges", []):
                     f = edge.get("file") or edge.get("target_file")
-                    if f:
-                        files.add(f)
-            return set(list(files)[:limit])
+                    if f and f not in seen:
+                        seen.add(f)
+                        files_ordered.append(f)
+            return set(files_ordered[:limit])
         return set()
-    except Exception:
+    except Exception as e:
+        print(f"  graph: parse failed for {symbol}: {e}", file=sys.stderr)
         return set()
 
 
@@ -151,7 +163,7 @@ def main():
         print(f"[{i + 1}/{len(tasks)}] {symbol} ({len(relevant)} ground-truth files)")
 
         for cond_name, runners in [
-            ("grep", lambda: run_grep(repo_path, symbol)),
+            ("grep", lambda: run_grep(repo_path, symbol, args.k)),
             ("spelunk_search", lambda: run_spelunk_search(repo_path, symbol, args.k)),
             ("spelunk_graph", lambda: run_spelunk_graph(repo_path, symbol, args.k)),
         ]:
--- a/bench/graph/evaluate.py
+++ b/bench/graph/evaluate.py
@@ -1,0 +1,213 @@
+#!/usr/bin/env python3
+"""Code-graph benchmark — grep vs spelunk_search vs spelunk_graph.
+
+For each task (a symbol in an indexed repo), runs three conditions and
+measures how well each retrieves the ground-truth set of files containing
+callers/callees/implementers of that symbol.
+
+Conditions:
+    grep           — git grep <symbol> over the repo
+    spelunk_search — spelunk search <symbol> (semantic)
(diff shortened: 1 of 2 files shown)


usercise (commented):
**Status: infrastructure is right shape, but `Closes #230` is premature — same scaffolding-vs-deliverable split as PR #236.**

The three-condition eval works, the metric trio (precision@k / recall@k / F1) is the right shape, and pure-retrieval (no LLM) is the right approach for this pillar. Two structural concerns and one premature-close.

## Premature `Closes #230` — 2 tasks, 1 repo, vs the issue's ≥30 tasks across ≥3 repos

`bench/graph/tasks.json` has 2 tasks, both pointing at `~/Projects/deepseek-spelunk` itself. Issue #230's acceptance asks for:

> ≥30 tasks across ≥3 repos with ground truth committed.

This PR delivers the eval framework (good) but not the task corpus (the bigger half of the work). Same pattern as PR #236 with #226 — infrastructure landed, content followed in #237.

**Recommended:** swap `Closes #230` → `Refs #230`, merge this as scaffolding, open a follow-up issue *\"bench(graph): author code-graph task corpus (≥30 tasks × ≥3 repos with reference-tool ground truth)\"*. Without ground truth from a reference tool (rust-analyzer, jedi, etc. — issue #230 named these specifically), the corpus is just hand-curated guesses and doesn't carry signal.

Ground truth construction is the hardest part — needs to be either (a) derived from a language server / static analyser, or (b) human-audited from a known reference implementation. Worth its own issue.

## Concern — apples-to-oranges retrieval-size

`run_grep` (line 46) returns **every file containing the literal token** — no `--max-count`, no slicing to `args.k`. Meanwhile `run_spelunk_search` and `run_spelunk_graph` are capped at `args.k` (default 10). So for any symbol with broad textual presence (common helper names, generic verbs like `process` / `run` / `parse`), grep returns hundreds of files and precision tanks while spelunk gets the precision win essentially by retrieving fewer things.

This favours spelunk in a way that makes the comparison unfair *in spelunk's direction* — opposite of what we'd want when the pillar 1 / pillar 2 narrative is already \"semantic over keyword wins\". A reviewer auditing this would correctly flag the metric as cherry-picked.

Fix — either (a) limit grep to top-k results (`git grep -l <symbol> | head -n k` — but how to pick which k?), or (b) report **unranked** retrieval-set metrics rather than precision@k. For a graph benchmark where the question is \"can the tool find all related files,\" recall is the more meaningful metric. Precision can be reported with the actual retrieval size as an explicit denominator.

If you keep precision@k, document in the result JSON which k each condition was bound by, and consider reporting `precision_full` alongside `precision_at_k`.

## Concern — `spelunk graph` JSON shape may not match the parser

`run_spelunk_graph` (line 81-102) parses results expecting:

```python
for edge in r.get("edges", []):
    f = edge.get("file") or edge.get("target_file")
```

This assumes either a single object with an `edges` array, or a list of such objects. I don't have visibility into what `spelunk graph <symbol> --format json` actually emits today — could be a flat list of edges, could be a tree, could be node-edge pairs. The current parsing has a `try/except Exception: return set()` that swallows mismatches silently.

Easy verification — run it once during PR review:

```bash
$ spelunk graph run_agent --format json | head -30
```

If the shape doesn't match, `spelunk_graph` will score 0 on every task without any indication. Add a `print(\"unexpected graph shape\", r, file=sys.stderr)` in the fallback path, OR pin the expected shape in the test corpus and assert it.

Side issue: `set(list(files)[:limit])` (line 99) is non-deterministic — `set → list` order is unspecified, and slicing arbitrary order produces a different top-k each run. Build a *list* in insertion order from the graph output, dedupe with `dict.fromkeys()`, then slice.

## Smaller concerns

- **No `scaffold_hash` / `seed` fields** in the output JSON. Even pure-retrieval benchmarks benefit from the reproducibility-contract fields for cross-PR comparison. Easy to add.
- **`run_spelunk_search` uses `r.get(\"file\", \"\")`** — verify this is the actual field name in `spelunk search --format json` output. (Spot-check while you're verifying #2's shape.)
- **`tasks.json` ground truth for `check_hit`** lists only `bench/memory/decision_archaeology.py`, but `check_hit` is a generic function name — grep would also legitimately match anywhere else the string appears. Ground truth should be \"files defining or relevantly using the symbol\", which for `check_hit` is just the one file — so this is technically right, but a comment in `tasks.json` documenting the convention would help future contributors.

## Suggested smoke test

Before merging, run the eval against its own 2-task corpus and verify:
1. All three conditions return non-empty sets (not silently zeroed).
2. `spelunk_graph` actually populates `files` (this is the field shape sanity check).
3. Precision numbers for grep aren't artificially low because retrieval is unbounded.

If those pass, the framework is sound and the follow-up is just dataset authoring.

usercise (commented):
**Status: ✅ all prior feedback addressed cleanly. Ready to merge from my side. (Would be `--approve` but GitHub blocks self-approval.)**

Every concern from the prior review is fixed:

- ✅ **`Closes #230` → `Refs #230`** with follow-up [#248](https://github.com/usercise/spelunk/issues/248) tracking the actual task corpus authoring. Same scaffolding-merge model used for #226/#236 and (recommended for) #228/#243 and #231/#244 — consistent and right.
- ✅ **`run_grep` apples-to-oranges retrieval-size fixed** — now `git grep -E "\\b<symbol>\\b"` for word-boundary matching, and `set(files[:limit])` caps grep at the same `k` as the other conditions. The comparison is fair: each tool gets the same retrieval budget.
- ✅ **`spelunk_graph` deterministic ordering** — `files_ordered` list with a `seen` dedupe set preserves insertion order from the graph output. Slicing produces a stable top-k.
- ✅ **Stderr warnings on unexpected shapes** — both \"unexpected shape\" (graph output without `edges`) and `except Exception as e: print(...)` (parse failures) now surface to stderr instead of silently returning empty. Same pattern that revealed the FTS5 syntax error in #235.

I also confirmed [#248](https://github.com/usercise/spelunk/issues/248) was filed correctly — \"≥30 find-all-callers tasks across ≥3 repos with ground-truth file sets\", with the rust-analyzer / jedi reference-tool note from the original #230 issue carried forward. Good follow-up tracking.

## What this unblocks

Once #248 authors the real task corpus, the rerun produces signal for the previously-unmeasured pillar 2 (code graph). Per the architect briefing, that's a secondary pillar now — but having any honest number for it is better than the report's previous \"infrastructure ready\" placeholder.

## Smoke test recommendation

Before merging, just to verify the parser changes don't break against the current `spelunk graph` output shape, run:

```bash
spelunk graph run_agent --format json
python bench/graph/evaluate.py --tasks bench/graph/tasks.json --k 10
```

Expect all three conditions to return non-empty sets for at least one task (`run_agent` exists in both `bench/agents/agent.py` and `bench/memory/cross_session_handoff.py`, so grep should hit both). If `spelunk_graph` returns empty *and* prints \"unexpected shape\" to stderr, the parser still needs adjustment — but at least the failure will be loud now instead of silent.

Otherwise — clean iteration, ship it.

PR #246, opened 2026-05-17
docs(bench): rewrite benchmarking report as infrastructure status

usercise:
Closes #232

Rewrite tmp/benchmarking-report.md:
- Remove all headline numbers from pre-control runs
- Executive summary: status table instead of claims
- Each section documents conditions, scripts, and reproduction commands
- Add Limitations section
- Add Scaffolding section for infrastructure-ready scripts
- Mark stale baselines

### DIFF
--- a/tmp/benchmarking-report.md
+++ b/tmp/benchmarking-report.md
@@ -1,0 +1,206 @@
+# spelunk Benchmarking Report
+
+**Date:** 2026-05-17
+**Model:** `deepseek-v4-flash` (DeepSeek V4 Flash, non-thinking)
+**spelunk version:** 0.6.0
+**Status:** Infrastructure complete. Controlled reruns pending.
+
+---
+
+## Executive Summary
+
+We built a reproducible benchmarking framework covering spelunk's three
+value pillars. All benchmark scripts produce verifiable, seed-controlled
+output with reproducibility-contract fields. Controlled experiments with
+matched compute budgets, n>=30, and correctness checks are now ready to
+run — the framework is in place; the headline numbers await the rerun.
+
+| Pillar | Benchmark | Status |
+|--------|-----------|--------|
+| Semantic search | RepoBench cross-file completion | 4-condition eval ready; pending controlled rerun with indexed-repo overlap |
+| Project memory | Decision archaeology | 4-condition eval ready; pending blind question sets |
+| Project memory | Cross-session handoff | 3-condition eval with verify_cmd ready; pending task corpus |
+| Code graph | Call-graph retrieval | 3-condition eval ready; pending task corpus |
+| Performance | Index/search/memory at scale | `perf_scale.sh` orchestrator ready; pending scale runs |
+
+---
+
+## 1. RepoBench — Cross-File Code Completion
+
+**Dataset:** [RepoBench-Python v1.1](https://huggingface.co/datasets/tianyang/repobench_python_v1.1),
+`cross_file_first` split.
+
+**Four conditions** (matched 5-turn compute budget):
+
+| Condition | Loop | Tools | Isolates |
+|-----------|------|-------|----------|
+| `baseline_single_shot` | No | None | Single-call baseline |
+| `multi_turn_no_tools` | Yes | None | Loop without retrieval |
+| `naive_search` | Yes | `read_file`, `run_grep` | Tool-calling without semantic search |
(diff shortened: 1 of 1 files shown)


usercise (commented):
**Status: this is the report we needed. Reads honestly, drops every unsupported headline, and the Limitations section calls out the right things. Approve from my side once a few cross-PR-reference issues are sorted (would be `--approve` if GitHub allowed self-approval).**

The rewrite does exactly what #232 asked for:

- ✅ No headline numbers cited anywhere — `4×`, `7.5×`, `100%`, `0%` are all gone.
- ✅ Executive summary is a status table (\"4-condition eval ready; pending controlled rerun\") rather than a results table.
- ✅ Each pillar documents conditions, scripts, reproduction commands — and not results.
- ✅ Limitations section is explicit and honest (API non-determinism, repo availability, indexed-repo overlap, blind question sets, single-machine perf, handoff task corpus).
- ✅ Scaffolding section moved out of the executive summary, where it was misleading before.
- ✅ Stale baseline marked.

This is the right tone for what we have. A founder/brand-manager could now cite this without reservation — \"infrastructure for controlled comparisons is built; the controlled comparisons themselves are the next milestone.\"

## Concerns — forward references to open PRs

Section 3 (Cross-Session Handoff) describes features that I believe live in PR [#243](https://github.com/usercise/spelunk/pull/243) (still open): `verify_cmd`, `--session-1-turns`, `handoff_tasks.json`, the three Session 2 conditions. Section 5 (SWE-bench) describes [#242](https://github.com/usercise/spelunk/pull/242)'s `--save-patch`, `export_patches.py`, `swebench_eval.sh`. Section 6 (Performance) describes [#244](https://github.com/usercise/spelunk/pull/244)'s `perf_scale.sh`. Section 4 (Code Graph) describes [#245](https://github.com/usercise/spelunk/pull/245)'s `bench/graph/evaluate.py`.

Issue #232 anticipated this — it explicitly said:

> Depends on #224–#231 (in practice, ones that ship; others get de-scoped explicitly in the report).

Two reasonable paths:

- **A:** hold this PR until #242 / #243 / #244 / #245 all land. Cleanest narrative, no risk of dangling references. Cost: this PR sits in review queue for however long the last of those takes.
- **B:** merge this now with an upfront note (\"References to scripts in sections 3–6 land via PRs #242–#245; if any are de-scoped, this report's matching section is updated in a follow-up\"). Acceptable, but every section currently reads as if the script is on `main`.

I'd lean A — the report is the wrap-up artefact, it should reflect what's actually shipped. But if there's external pressure to publish status, B with a clear note also works.

Side note: my reviews on #243 / #244 / #245 each have outstanding concerns. None of them are blockers in spirit but they need follow-up. Worth confirming those are addressed before this report cites them as ready.

## Smaller suggestions

**Link back to the review trail.** The report is missing context about *why* this exists — i.e., the original benchmark review that triggered the entire fix-plan workstream. A one-line backreference at the top makes future readers less confused:

> *See `tmp/benchmark-fix-plan.md` and `~/Documents/Spelunk/background/from-benchmark-review.md` for the context that drove this rewrite.*

**Make the 24/50 SWE-bench denominator more explicit.** The Limitations bullet:

> The SWE-bench Verified dataset contains 24 of the 50 pinned tasks. The remaining 26 require the full SWE-bench dataset.

Reads as a dataset-availability note. Better wording — call out the consequence:

> Any `resolve_rate` figure produced by `swebench_eval.sh` until the 26 missing repos are cloned is computed on 24 of 50 tasks. The result JSON must include `tasks_evaluated` and `tasks_total` distinct from `resolve_rate` so a reader can't conflate \"X% on Verified\" with \"X% on 50/50\".

This is the same overstatement risk that prompted the original benchmark review.

**Link #237 from the Decision Archaeology section.** Status row says \"pending blind question sets\" — link to [#237](https://github.com/usercise/spelunk/issues/237) explicitly so readers can track it.

**\"Pending task corpus\" rows.** Sections 3 (Handoff) and 4 (Graph) both say \"pending task corpus\". Worth filing follow-up issues for these (sister to #237) and linking them — otherwise these become orphaned TODOs.

## What I'd like to see before final approval

Just a few lines added to the report (the cross-references and the explicit SWE-bench denominator note) and a decision on path A vs B for the open-PR dependencies. The structure and content of the rewrite are exactly right.

Great job on the tone — \"pending controlled rerun\" is honest in a way that protects future credibility without losing the \"we've done the work\" framing.

PR #278, opened 2026-05-24
feat(#269): spelunk status --format json

### LINKED ISSUE #269: spelunk status --format json
## Outcome
`spelunk status` accepts `--format json` and emits a stable, documented schema for agents and scripts to consume.

## Acceptance
- [ ] Flag parses; default output remains human-readable text
- [ ] JSON schema documented (in `docs/` or inline)
- [ ] Stable field names — breaking changes require deprecation
- [ ] Test coverage for the JSON shape

## Context
WS-1 (OSS v1.0) must-ship.

usercise:
## Summary

- Adds `--format json` flag to `spelunk status` for machine-readable output
- Stable schema fields:
  - `version` — crate version string
  - `project` — absolute path to the indexed project root
  - `db_path` — path to the SQLite index
  - `indexed_files` / `file_count` — number of source files in the index (both keys present for compat)
  - `total_chunks` — total AST chunks indexed
  - `languages` — `[{name, file_count}]` sorted by count descending
  - `embedding_dim` — embedding vector dimension (null if no semantic search)
  - `has_semantic_search` — boolean
  - `last_indexed_at` — ISO-8601 UTC timestamp of most recent index run (null if never)
  - `memory_entries` — count of memory entries
- Extension fields preserved (tier, capabilities, drift_candidates, usage_7d)
- New `Database::language_stats()` query + `LanguageStat` struct in spelunk-core
- New e2e test `test_status_json_stable_schema` verifying all stable keys; pre-existing `status_json_file_count_matches_ls_files_count` continues to pass

## Test plan

- [x] `cargo test -p spelunk-cli` — all 95 tests pass (0 failures)
- [x] `cargo clippy -- -D warnings` — clean
- [x] `cargo fmt --check` — clean (pre-commit hook verified)
- [ ] Manual: `spelunk status --format json | jq .` on an indexed project
- [ ] Manual: `spelunk status` (no flag) — verify text output unchanged

Closes #269


crates/spelunk-cli/src/cli/cmd/status.rs
```diff
@@ -24,15 +24,58 @@ use crate::{
...
+                .ok()
+                .and_then(|cwd| r.find_project_for_path(&cwd).ok().flatten())
+        });
+        let db_path = match &project_entry {
+            Some(p) => p.db_path.clone(),
+            None => {
+                let (p, _) = resolve_project_and_deps(None, &cfg)?;
+                p
+            }
+        };
```
usercise: Surely this is core code that we should reuse. Every cmd that talks to the DB needs to know how to resolve the project.

crates/spelunk-cli/src/cli/cmd/status.rs
```diff
@@ -285,6 +379,38 @@ fn print_tier_section(tier: &Tier, cfg: &Config) {
     println!();
 }
 
+/// Convert a Unix timestamp (seconds) to a `(date, time)` tuple formatted as
+/// `"YYYY-MM-DD"` and `"HH:MM:SS"` (UTC), without pulling in an external date
+/// library.
+///
+/// Returns `("1970-01-01", "00:00:00")` for `ts = 0`.
+fn iso8601_from_unix(ts: u64) -> (String, String) {
```
usercise: Let's use chrono crate for datetime & time functions. Please review the rest of the codebase to make sure we don't have more of these local time functions.

crates/spelunk-cli/src/cli/cmd/status.rs
usercise: Done — hoisted into `spelunk_core::registry::resolve_project_context` (returns `ResolvedProject { project, db_path, deps }`). `search::resolve_project_and_deps` is now a thin wrapper that adds existence-check bails, and `status` uses the core helper directly for both JSON and text branches. Net effect: status.rs lost ~50 lines of duplicate registry lookup.

crates/spelunk-cli/src/cli/cmd/status.rs
usercise: Done — added `chrono` to workspace deps (lean default-features: `clock`, `serde`, `std` only). Replacements:

- `status::iso8601_from_unix` → `DateTime::<Utc>::from_timestamp().format("%Y-%m-%dT%H:%M:%SZ")`
- `utils::dates::parse_iso8601_to_epoch` / `days_since_epoch` → `NaiveDate(Time)::parse_from_str` (also tightened the accepted format docs)
- `utils::format_age` + `status::format_age` → `chrono::Utc::now`

Audit: those were the only hand-rolled calendar functions. The remaining `SystemTime::now().duration_since(UNIX_EPOCH).as_secs()` spots (a handful in storage/server) are just "current unix seconds" — no calendar arithmetic — so I left them alone. Happy to swap them too if you'd like consistency.

PR #326, opened 2026-06-03
feat(capability): loopback auto-discovery — spelunk#316 (0.8.0 cold-start Section A)

### LINKED ISSUE #316: feat(#303-A): loopback auto-discovery + SPELUNK_NO_SERVER
## Parent
Tracked by #303 (0.8.0 server-default cold-start).

## Summary
Implement loopback auto-discovery in `crates/spelunk-cli/src/capability.rs` and `crates/spelunk-core/src/config.rs`.

## Spec (Section A from #303)

When `cfg.server_url` is `None`:
1. Read `~/.local/state/spelunk/server.port` if it exists; use `http://127.0.0.1:<port>`.
2. Otherwise default-probe `http://127.0.0.1:7777` with a 250 ms timeout.
3. On success, treat as `Tier::Server` for the rest of the process.
4. On failure, fall through to bootstrap policy (Section B / #304).

Additional:
- `Config::validate()` no longer requires `project_id` when `server_url` is a loopback address (loopback derivation handles it — see #307).
- `SPELUNK_NO_SERVER=1` short-circuits the probe and forces `Tier::Offline`.

## Definition of done
- [ ] `probe()` in `capability.rs` reads the port file and falls back to port 7777
- [ ] 250 ms timeout on the loopback probe (distinct from the current 2 s timeout used for explicit remote URLs)
- [ ] `SPELUNK_NO_SERVER=1` → `Tier::Offline` without probing
- [ ] `Config::validate()` exempts loopback URLs from the `project_id` requirement
- [ ] Unit tests for all branches (port file exists, port file absent, no-server env var, loopback URL validation bypass)

usercise:
## Summary

Implements Section A of the 0.8.0 server-default cold-start spec (spelunk#303, sub-issue #316).

**What this does:**
- `spelunk-cli/src/capability.rs`: when `cfg.server_url` is `None`, reads `~/.local/state/spelunk/server.port` (written by `spelunk server start`, coming in #317); probes that port or falls back to `127.0.0.1:7777` with a **250 ms** timeout; respects `SPELUNK_NO_SERVER=1|true|yes` env var; adds `auto_discovered: bool` to `Tier::Server` for future `spelunk status` UX annotation (#324)
- `spelunk-core/src/config.rs`: new `pub is_loopback_url()` helper; `Config::validate()` exempts loopback URLs from the `project_id` requirement; full unit tests for all branches
- `spelunk-cli/src/cli/cmd/check.rs` + `status.rs`: pattern matches updated for new enum field

**Sub-issues filed:** #316–#325 (full cold-start decomposition)

## Test plan
- [x] `cargo fmt && cargo clippy -- -D warnings` — clean
- [x] `cargo test` — all passing (59 spelunk-core + 44+ spelunk-cli)
- [x] No new dependencies (Cargo.lock unchanged)

## Dependencies
- Depends on: #317 (`spelunk server start`) to write `server.port` state file
- Unblocks: #318 (init auto-spawn), #323 (search Tier-0 fall-through)

Closes #316. Part of #303.

### DIFF
--- a/crates/spelunk-cli/src/capability.rs
+++ b/crates/spelunk-cli/src/capability.rs
@@ -26,9 +26,24 @@ use tokio::sync::OnceCell;
 
 use crate::config::Config;
 
-/// State file directory: `~/.local/state/spelunk/`
+/// State file directory: `~/.local/state/spelunk/`.
+///
+/// On all platforms we use `~/.local/state` rather than the OS-native state dir.
+/// This mirrors the deliberate choice made for the config dir
+/// (`spelunk_config_dir` in spelunk-core's `config.rs`, which uses `~/.config`
+/// on every platform): it keeps the path identical across Linux and macOS, and
+/// matches what the CLI documentation and error messages say.
+///
+/// It also sidesteps a concrete portability bug: `dirs::state_dir()` returns
+/// `None` on macOS (dirs v6 has no XDG_STATE_HOME equivalent there), which
+/// silently disabled loopback auto-discovery on the primary dev platform
+/// (spelunk#316). Returns `None` only when the home directory can't be resolved.
+///
+/// NOTE for spelunk#317 (writer side, `spelunk server start`): the writer MUST
+/// write `server.port` into this exact directory so reader and writer agree.
+/// Use the same `~/.local/state/spelunk/` path on every platform.
 fn spelunk_state_dir() -> Option<std::path::PathBuf> {
-    dirs::state_dir().map(|d| d.join("spelunk"))
+    dirs::home_dir().map(|home| home.join(".local").join("state").join("spelunk"))
 }
 
 /// Read the port written by `spelunk server start` into
@@ -567,5 +582,28 @@ mod tests {
     #[test]
     fn default_loopback_port_is_7777() {
         assert_eq!(DEFAULT_LOOPBACK_PORT, 7777);
+    }
+
+    // ── SPELUNK_NO_SERVER short-circuit behaviour ─────────────────────────────
+    //
+    // These tests mutate the process-global `SPELUNK_NO_SERVER` env var, so they
+    // are serialised against each other to avoid cross-test interference.
+
+    #[tokio::test]
+    #[serial_test::serial(spelunk_no_server_env)]
+    async fn spelunk_no_server_forces_offline() {
+        // SAFETY: serialised via #[serial] so no other test reads/writes this
+        // env var concurrently; restored before the guard scope ends.
+        for val in ["1", "true", "yes"] {
+            unsafe { std::env::set_var("SPELUNK_NO_SERVER", val) };
+            // server_url = None so that, absent the short-circuit, the probe would
+            // attempt loopback auto-discovery; the short-circuit must win.
+            let tier = probe(None, None).await;
+            assert!(
+                matches!(tier, Tier::Offline),
+                "SPELUNK_NO_SERVER={val} should force Tier::Offline, got {tier:?}"
+            );
+        }
+        unsafe { std::env::remove_var("SPELUNK_NO_SERVER") };
     }
 }
--- a/crates/spelunk-cli/src/capability.rs
+++ b/crates/spelunk-cli/src/capability.rs
@@ -3,6 +3,21 @@
 //! Tier 0 (Offline): no server_url configured, or server unreachable.
 //! Tier 1 (Server):  server_url set and GET /v1/health succeeds.
 //!
+//! ## Loopback auto-discovery (spelunk#316 / 0.8.0)
+//!
+//! When `cfg.server_url` is `None` **and** `SPELUNK_NO_SERVER` is not set, the probe
+//! attempts to reach a locally-running spelunk-server before falling through to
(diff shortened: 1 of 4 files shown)


usercise (commented):
## QA Review — PR #326 (feat(capability): loopback auto-discovery, spelunk#316)

Reviewed against spelunk#303 Section A spec and sub-issue #316. Overall the core logic is sound and the main path is correct, but I found **two bugs** (one functional regression against the spec, one platform portability gap) plus advisory notes.

---

### BUG 1 (functional) — `spelunk check` still gates on `cfg.server_url.is_some()` — auto-discovered servers never shown

**File:** `crates/spelunk-cli/src/cli/cmd/check.rs`, line 117

```rust
if (effective == "text" || effective == "porcelain") && cfg.server_url.is_some() {
```

The spec (spelunk#303 Section F) explicitly requires:

> `spelunk check` shows server status **even when `server_url` was auto-discovered** (currently the line only prints when `cfg.server_url.is_some()`)

This guard is unchanged from main. When a user has no `server_url` in config and the auto-discovery probe succeeds, `spelunk check` will silently omit the server line entirely — contradicting the spec and leaving users unable to verify whether auto-discovery worked.

The guard should become something like `&& (cfg.server_url.is_some() || tier.is_server())`, evaluated after calling `get_tier()`. This is a **blocker** for the cold-start UX story.

---

### BUG 2 (portability) — `dirs::state_dir()` returns `None` on macOS; server.port file is silently skipped

**File:** `crates/spelunk-cli/src/capability.rs`, `spelunk_state_dir()` (new function, ~line 30)

```rust
fn spelunk_state_dir() -> Option<std::path::PathBuf> {
    dirs::state_dir().map(|d| d.join("spelunk"))
}
```

`dirs` v6 (confirmed in `Cargo.lock`: `dirs 6.0.0`) returns `None` for `state_dir()` on macOS — macOS has no XDG state directory equivalent. The `?` in `read_server_port_file()` propagates this to `None`, so the port file read is silently skipped on macOS. The fallback probe to port 7777 still runs, but any non-default port set by `spelunk server start` is invisible on macOS.

`spelunk-server` (sub-issue #317) will presumably also use `state_dir()` to write the file; if it encounters the same `None`, the port file is never written either, making this a consistent no-op rather than a broken read — but it's fragile and macOS users on non-default ports will fall through to the default 7777 probe regardless of what port the server actually started on.

Recommended fix: add a macOS fallback, e.g. `dirs::data_local_dir()` or a hardcoded `~/.local/state/spelunk` via `dirs::home_dir()`. The XDG spec path is fine on Linux; just need a macOS alternative. Coordinate with #317 so both reader and writer agree on the same path.

---

### Advisory — No unit test exercises `SPELUNK_NO_SERVER` env var end-to-end

The spec requires `SPELUNK_NO_SERVER=1` to force `Tier::Offline`. There are three tests that verify the constants (`loopback_probe_timeout_is_250ms`, `remote_probe_timeout_is_2s`, `default_loopback_port_is_7777`) but no test calls `probe()` with `SPELUNK_NO_SERVER` set in the environment and asserts `Tier::Offline` is returned. Given `probe()` is `async` and the matching `Ok("1") | Ok("true") | Ok("yes")` is non-trivial, a behavioral test would catch any future regression here. The existing `read_server_port_file_returns_none_when_absent` test similarly just calls the function without asserting on state-dir behavior (with a comment deferring to integration tests). Acceptable for now, but both branches deserve integration coverage before the full cold-start stack is wired.

---

### Advisory — `SPELUNK_NO_SERVER` does not short-circuit when `server_url` is explicitly set

When a user has `server_url` set in config *and* sets `SPELUNK_NO_SERVER=1`, the env var wins (returns `Tier::Offline` at step 1 before step 2 runs). This is correct behavior, but there is no comment or test documenting this interaction. Worth a note in the doc comment on `probe()` since it's a non-obvious precedence.

---

### Checklist (what's correct)

- [x] 250 ms loopback probe timeout — matches spec exactly (`LOOPBACK_PROBE_TIMEOUT = Duration::from_millis(250)`)
- [x] 2 s remote probe timeout preserved for explicit `server_url`
- [x] `SPELUNK_NO_SERVER` accepts `"1"`, `"true"`, `"yes"` — reasonable superset of spec's `=1`
- [x] `auto_discovered: bool` field added correctly to `Tier::Server`
- [x] All three existing `Tier::Server { url, caps }` pattern matches in `check.rs` and `status.rs` updated with `..` — no unhandled struct field compile errors
- [x] `is_loopback_url()` correctly handles `localhost`, `127.x.x.x`, `[::1]` including IPv6 bracket form — tests are thorough
- [x] `Config::validate()` loopback exemption logic and tests are correct
- [x] No new dependencies (Cargo.lock unchanged — confirmed)
- [x] No credentials or tokens introduced
- [x] No SQL string formatting

---

**Summary:** Bug 1 (`spelunk check` guard) is a spec violation that should block merge — the `cfg.server_url.is_some()` guard must be removed/relaxed to include the auto-discovered case. Bug 2 (macOS `state_dir`) is a platform correctness issue that will silently affect all macOS users on non-default ports; should be fixed before #317 is merged so both sides agree on the state path.

PR #338, opened 2026-06-05
feat(#325): install paths — install.sh, deb packaging, service units, brew tap

usercise:
Closes #325

## Summary

- **`install.sh`** — POSIX `curl | sh` installer: detects OS/arch (Linux x86_64/aarch64, macOS x86_64/arm64), maps to the correct release tarball, installs to `/usr/local/bin` (falls back to `~/.local/bin` if not writable), supports `--dry-run`, prints `spelunk --version` on success.
- **`packaging/spelunk-server.plist`** — macOS LaunchAgent (`cloud.spelunk.server`) that runs `spelunk-server --port 7777` at login, with `KeepAlive=true`, stdout/stderr to `/tmp/spelunk-server.{log,err}`.
- **`packaging/spelunk-server.service`** — Linux systemd user unit (`Type=simple`, `Restart=on-failure`, `WantedBy=default.target`).
- **`.github/workflows/release.yml`** — new `deb` job (runs after `build`, before `release`) that assembles a proper `DEBIAN/control` + binary layout and calls `dpkg-deb --build`; the `release` job now also uploads `*.deb` assets.
- **`homebrew-tap/`** — `Formula/spelunk.rb` and `.github/workflows/update-formula.yml` (workflow\_dispatch with version/url/sha256 inputs) ready to push to `spelunk-cloud/homebrew-spelunk`. Run `scripts/push-homebrew-tap.sh` once to create the tap repo and push.

## Test plan

- `bash install.sh --dry-run` prints the plan (OS, arch, target, download URL, install path) and exits 0 without touching the filesystem.
- YAML files are syntactically valid (`yamllint` / GitHub Actions linter).
- deb job: on next tag push the CI matrix will produce `spelunk_VERSION_amd64.deb`; install and verify `dpkg -l spelunk`, `spelunk --version`, `systemctl --user enable spelunk-server`.
- Homebrew tap: after running `scripts/push-homebrew-tap.sh`, `brew tap spelunk-cloud/spelunk && brew install spelunk` installs both binaries (sha256 needs updating for the actual v0.8.0 release).

## Follow-up

- Update `sha256` in `homebrew-tap/Formula/spelunk.rb` once v0.8.0 is tagged and the macOS tarball is published.
- Add an `install.sh` download link to the README.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #345, opened 2026-06-06
feat: move Homebrew formula in-repo (ripgrep model), fix license and 404

usercise:
## Summary

- Adds `HomebrewFormula/spelunk.rb` at the repo root, following the ripgrep in-repo tap model
- Install via `brew tap spelunk-cloud/spelunk https://github.com/spelunk-cloud/spelunk` then `brew install spelunk-cloud/spelunk/spelunk`
- Deletes the separate `homebrew-tap/` directory and `scripts/push-homebrew-tap.sh`

## Fixes

- **404 (structural fix):** Uses `version "0.8.0"` + `#{version}` interpolation so each platform URL resolves the correct versioned asset (e.g. `spelunk-v0.8.0-aarch64-apple-darwin.tar.gz`)
- **Wrong license:** EUPL-1.2 → MIT
- **Multi-platform:** Covers macOS arm64, macOS x86_64, Linux arm64, Linux x86_64 using Homebrew DSL (`on_macos`/`on_linux` + `on_arm`/`on_intel`)

## sha256 automation

Adds an `update-homebrew-formula` job to `release.yml` that runs after each stable tag release:
1. Downloads all four platform tarballs from the just-created GitHub release
2. Computes `sha256` for each
3. Patches `HomebrewFormula/spelunk.rb` and commits back to `main`

The formula ships with `TODO_*` placeholder sha256 values; these will be replaced on the next tag push.

## Manual update instructions

If you need to update the formula by hand (e.g. before CI is wired up):
```sh
shasum -a 256 spelunk-v<VERSION>-aarch64-apple-darwin.tar.gz
shasum -a 256 spelunk-v<VERSION>-x86_64-apple-darwin.tar.gz
shasum -a 256 spelunk-v<VERSION>-aarch64-unknown-linux-gnu.tar.gz
shasum -a 256 spelunk-v<VERSION>-x86_64-unknown-linux-gnu.tar.gz
# Replace the corresponding sha256 values in HomebrewFormula/spelunk.rb
```

## Test plan

- [ ] Verify `brew tap spelunk-cloud/spelunk https://github.com/spelunk-cloud/spelunk` succeeds after merge
- [ ] Verify `brew install spelunk-cloud/spelunk/spelunk` resolves the correct tarball URL
- [ ] Push a test tag and confirm the `update-homebrew-formula` CI job patches the sha256 values correctly
- [ ] `brew audit --strict HomebrewFormula/spelunk.rb` passes (may need real sha256 values first)

Closes #342, #338, #340


.github/workflows/release.yml
```diff
@@ -268,3 +268,94 @@ EOF
...
+
+          cat "$FORMULA"
+
+      - name: Commit and push updated formula
+        run: |
+          git config user.name  "github-actions[bot]"
+          git config user.email "github-actions[bot]@users.noreply.github.com"
+          git add HomebrewFormula/spelunk.rb
+          git commit -m "chore: update Homebrew formula to v${VER}" || echo "No changes to commit"
+          git push origin main
```
usercise: `main` direct changes are blocked, this would need to be a PR.

.github/workflows/release.yml
```diff
@@ -268,3 +268,94 @@ EOF
...
+              pattern = (
+                  r'(url "https://github\.com/spelunk-cloud/spelunk/releases/download/v[^/]+/'
+                  + re.escape(f'spelunk-v') + r'[^"]*-' + re.escape(target) + r'\.tar\.gz"'
+                  + r'\s+sha256 ")[0-9a-f]+(")'
+              )
+              content = re.sub(pattern, r'\g<1>' + sha + r'\g<2>', content)
+
+          with open(formula_path, "w") as f:
+              f.write(content)
+          PYEOF
```
usercise: This should be split out into a script file in `.github/scripts`. Can we change this script to javascript so its easier to maintain. Also can we change this to instead of using string replacement use a string literal template and just print the file we want. Then there should be no reason to have a different version for the first run either.

.github/workflows/release.yml
```diff
@@ -268,3 +268,94 @@ EOF
...
+          SHA_AARCH64_DARWIN="$(download_sha256 aarch64-apple-darwin)"
+          SHA_X86_64_DARWIN="$(download_sha256 x86_64-apple-darwin)"
+          SHA_AARCH64_LINUX="$(download_sha256 aarch64-unknown-linux-gnu)"
+          SHA_X86_64_LINUX="$(download_sha256 x86_64-unknown-linux-gnu)"
+
+          echo "SHA_AARCH64_DARWIN=${SHA_AARCH64_DARWIN}" >> "$GITHUB_ENV"
+          echo "SHA_X86_64_DARWIN=${SHA_X86_64_DARWIN}"   >> "$GITHUB_ENV"
+          echo "SHA_AARCH64_LINUX=${SHA_AARCH64_LINUX}"   >> "$GITHUB_ENV"
+          echo "SHA_X86_64_LINUX=${SHA_X86_64_LINUX}"     >> "$GITHUB_ENV"
+          echo "VER=${VER}"                                >> "$GITHUB_ENV"
```
usercise: Can we not read this from the github release? It seems to me that `gh release view --json ${VER}` should give us the SHA256 of each release version.

.github/workflows/release.yml
```diff
@@ -268,3 +268,94 @@ EOF
...
+          VER="${VERSION#v}"
+
+          download_sha256() {
+            local target="$1"
+            local url="https://github.com/spelunk-cloud/spelunk/releases/download/${VERSION}/spelunk-${VERSION}-${target}.tar.gz"
+            curl -fsSL "$url" | sha256sum | awk '{print $1}'
+          }
+
+          SHA_AARCH64_DARWIN="$(download_sha256 aarch64-apple-darwin)"
+          SHA_X86_64_DARWIN="$(download_sha256 x86_64-apple-darwin)"
```
usercise: I think we can safely drop X86_64_DARWIN as a prebuilt version, we can leave instructions in the doc for building yourself. Apple has of today deprecated this arch and it is no longer supported by them, it was replaced 6 years ago and most engineers will have newer machines.

.github/workflows/release.yml
usercise: Fixed in 678de07 — the job now creates a `chore/homebrew-formula-v<VER>` branch, commits the regenerated formula there, and runs `gh pr create` to open a PR against `main` instead of pushing directly (which would be rejected by branch protection anyway).

.github/workflows/release.yml
usercise: Done in 678de07 — extracted into `.github/scripts/update-homebrew-formula.js` (Node, builtins only — `fs`/`path`, no extra npm deps). It builds the entire `HomebrewFormula/spelunk.rb` from a string-literal template (template literal) and writes the file outright, replacing the sed + python-regex patching. Since the file is always regenerated from scratch, there is no first-run-vs-subsequent-run distinction and no `TODO_*` placeholder matching at all — `release.yml` just calls `node .github/scripts/update-homebrew-formula.js --version … --sha-* …`. The committed `HomebrewFormula/spelunk.rb` was regenerated from the template (using v0.7.1's real release digests) so the two stay consistent — verified `node --check` + ran the script with sample values to confirm valid output, plus a Ruby syntax check on the generated formula.

.github/workflows/release.yml
usercise: Done in 678de07 — replaced `download_sha256()` (curl + sha256sum) with `gh release view "${VERSION}" --json assets --jq ".assets[] | select(.name == \"<asset>\") | .digest"`, stripping the `sha256:` prefix. Each of the three kept target tarball names is mapped to its digest under `set -euo pipefail`, so a missing/empty digest fails the job loudly rather than silently shipping a bad sha256.

.github/workflows/release.yml
usercise: Done in 678de07 — removed `x86_64-apple-darwin` from both build-matrix variants (default and `rich-formats`) and from `HomebrewFormula/spelunk.rb` (kept macOS arm64, Linux arm64, Linux x86_64). This also meant removing the `universal-macos` job: a "universal" `lipo` binary requires both x86_64 and aarch64 darwin slices, so it can no longer be built once x86_64 is dropped (and `release` no longer `needs` it). Updated `install.sh` to fail with a clear build-from-source pointer for `macos-x86_64` instead of resolving a tarball that no longer exists, and added a build-from-source note to `docs/getting-started.md` (linking to `docs/building.md`, which already documents the generic `cargo build --release` path) plus an updated platform table/note in `docs/releasing.md`.

PR #421, opened 2026-06-16
fix(server): use time-ordered UUIDv7 for instance_id; rename misnamed CLI nonce helper (#416)

usercise:
## What

- **Server `instance_id` now uses UUIDv7 instead of a hand-rolled v4.** `get_or_create_instance_id` still draws 16 bytes of entropy from SQLite `randomblob(16)`, but the new `format_uuid_v7(hex, unix_ts_ms)` packs the millisecond Unix timestamp into the high 48 bits, sets the version nibble to `7`, and sets the RFC-4122 variant bits to `10`. Result: instance IDs are time-ordered (lexicographic by leading bytes) while remaining standard 36-char UUID strings.
- **Renamed the misnamed `uuid_v4_hex()` -> `query_nonce_hex()`** in `spelunk-cli/src/server_client.rs`. That helper never produced an RFC-compliant UUID; it builds a cheap time+pid nonce for synthetic `query:<nonce>` chunk IDs. The rename plus a clarifying doc comment removes the misleading name; the call site in `embed_text` is updated.
- **Fixed inaccurate "UUID v4" doc comments** in `handlers.rs`, `lib.rs`, `docs/openapi.json`, and `docs/architecture/capability-tiers.md`.
- Regenerated `docs/openapi.json` (the `instance_id` description now reads UUID v7). Confirmed via `write_openapi_snapshot` that the committed file is the exact regenerated artifact (no drift).
- CHANGELOG `### Changed` entries added.

## Why

The previous code labeled the instance ID "UUID v4" but rolled the version/variant bits by hand. UUIDv7 gives the same uniqueness guarantee plus time-ordering, which is useful for restart-detection and log correlation. Correctness note worth flagging: the first cut of the variant-nibble mapping used ASCII-adjacency arithmetic and a `b'8'..=b'b'` range check; because `'9'` (0x39) and `'a'` (0x61) are not contiguous in ASCII, ~half of entropy values emitted `:` / `;` (0x3A/0x3B) -> non-canonical, invalid UUIDs. The testing stage caught this; the fix derives the variant via a direct `b\"89ab\"[entropy & 0b11]` lookup into the lowercase-hex set, and the regression test sweeps all 16 entropy values and asserts membership in `{8,9,a,b}` explicitly (not a byte range).

## Test coverage

- 6 unit tests on `format_uuid_v7`: version=7, variant=10xx, canonical all-lowercase-hex output, timestamp encoded in high 48 bits, lexicographic time-ordering, distinct-entropy non-collision, and a deterministic 0..16 sweep of the variant position.
- e2e `get_or_create_instance_id_is_stable_and_canonical` against an in-memory `ServerDb`: returned ID is canonical UUIDv7 and the persisted random tail is stable across calls.
- `query_nonce_hex_is_nonempty_lowercase_hex` guards the renamed helper's contract.

## Test plan

All run in the worktree against this branch:

- `cargo build` — clean.
- `cargo test -p spelunk-server` — 32 unit + 12 integration pass; 0 failed.
- `cargo test -p spelunk-cli` — all pass (incl. `query_nonce_hex_is_nonempty_lowercase_hex`); 0 failed.
- `cargo test -p spelunk-server format_uuid_v7` looped 5x — 6/6 pass every run (the original bug was probabilistic; the sweep test is deterministic over all 16 entropy values).
- `cargo clippy --all-targets` — no warnings, no errors.
- `cargo test -p spelunk-server write_openapi_snapshot` then `git status docs/openapi.json` — no drift; committed spec is the regenerated artifact.
- Stale-reference sweep: no remaining inaccurate "UUID v4" docs; no lingering `uuid_v4_hex` call sites (only the intentional historical mention in a doc comment + CHANGELOG).

Closes #416


crates/spelunk-cli/src/server_client.rs
```diff
@@ -262,7 +262,7 @@ impl ServerInferenceClient {
     /// The `chunk_id` is prefixed `query:` per ADR-002 so it is trivially
     /// distinguishable from real chunk ids in server logs.
     pub async fn embed_text(&self, text: &str) -> Result<Vec<f32>> {
-        let chunk_id = format!("query:{}", uuid_v4_hex());
+        let chunk_id = format!("query:{}", query_nonce_hex());
```
usercise:
```suggestion
        let chunk_id = format!("query:{}", Uuid::now_v7());
```
I don't understand why we need to roll our own implementation here, why can't we use the uuid crate?

crates/spelunk-server/src/db.rs
```diff
@@ -99,7 +101,7 @@ impl ServerDb {
                 |row| row.get(0),
             )
             .context("reading instance_id")?;
-        Ok(format_uuid_v4(&raw))
+        Ok(format_uuid_v7(&raw, unix_ts_millis()))
```
usercise: Again why are we rolling our own implementation, can't we use the `uuid` crate?

crates/spelunk-cli/src/server_client.rs
usercise: Good call - switched to the `uuid` crate. This now uses `Uuid::now_v7()` (uuid 1.x with the `v7` feature enabled in the workspace). Removed the hand-rolled `query_nonce_hex` helper entirely.

crates/spelunk-server/src/db.rs
usercise: Done - `get_or_create_instance_id` now persists a `Uuid::now_v7()` value verbatim (uuid crate), so the whole id is a stable time-ordered v7 UUID rather than re-derived each call. Removed the bespoke `format_uuid_v7`/`unix_ts_millis` helpers and their tests; kept a stability + version-7 test.

Issue #183, opened 2026-04-30
Spike: make embedding optional in memory add when no model is configured


usercise:
## Context

Part of #181 (git notes memory backend spike).

The research note's goal is zero-infrastructure first value: a user with no embedding model running can do `spelunk memory add --backend git-notes`. But this is currently blocked by the command handler, not the backend.

`src/cli/cmd/memory/add.rs:43–51` unconditionally calls `ActiveEmbedder::load()` and embeds before touching storage. Even with a `GitNotesBackend`, this path fails if no model is running.

`NoteInput.embedding` is already `Option<Vec<u8>>` and both `LocalMemoryBackend` and the remote backend handle `None` gracefully. The handler just never produces `None`.

## Task

1. In `src/cli/cmd/memory/add.rs`, make the embed step conditional: if `cfg.api_base_url` is not set (or no model is configured), skip embedding and pass `embedding: None` to the backend.
2. Add a log warning when skipping: `"No embedding model configured — entry stored without vector; semantic search will not surface it."`
3. Confirm that `memory search` and `memory list` (non-semantic) still work correctly for entries stored without embeddings.
4. Add a test: `spelunk memory add` with a config that has no `api_base_url` set succeeds and the entry appears in `memory list`.

## Success criteria

- `spelunk memory add --kind decision --title "..." --body "..."` succeeds with zero model infrastructure configured
- The stored entry appears in `spelunk memory list --kind decision`
- `spelunk memory search "<query>"` either skips the unembedded entry gracefully or returns an informative error, not a panic

## Notes

- `cargo fmt && cargo clippy` before committing
- This is a prerequisite for the git-notes backend spike — without it, the backend swap does not deliver the zero-infra promise

usercise:
Implemented in commit `0c85e68`. 

**Changes:** `src/cli/cmd/memory/add.rs` — extracted a `try_embed()` async helper that wraps the embedder load + embed call. On any error it logs the warning from the issue spec and returns `None`. The `NoteInput.embedding` field was already `Option<Vec<u8>>` and all backends handle `None` gracefully.

**Tests:** `tests/integration_git_notes.rs` exercises `memory add` with `embedding: None` (the git-notes backend never calls the embedder, so every add is an implicit no-embedding add) — all 5 tests pass.

Closes the gating requirement for #184.

usercise:
Spike complete. Implemented and passing per architect review on #181. Follow-on GA work tracked in #188 and #189.

PR #425, opened 2026-06-19
ADR-037 P1: local↔cloud two-way memory sync (sync mode, UUID identity, text-only push)

usercise:
Implements **P1 of ADR-037** (D1 + D2-minus-CRDT + D3) — the Founder's cross-machine dogfooding unblock. Closes #424. **Do not merge — Johan merges.**

### D1 — sync-mode flag
- New `SyncMode` enum (`offline` / `local_first` / `cloud_first`) + `Config.mode` (`Option`, so the serde default is preserved).
- **Defaults preserve today's behaviour**: no `server_url` ⇒ `offline`; `server_url` present ⇒ `local_first`. `cloud_first` is documented as debug/override only.
- `SPELUNK_MODE` env override (invalid value = hard error); `SPELUNK_NO_SERVER=1` remains a hard kill-switch forcing `offline`.
- The resolved mode (via `Config::resolve_mode`) now drives memory-backend selection and gates the capability probe, replacing the implicit "is the server reachable" branch.
- Loopback **inference** auto-discovery is preserved for the defaulted-offline case — it is inference-only and never a memory store (ADR-004); only an *explicit* offline (config/env/kill-switch) skips probing.

### D2 — real two-way sync (minus full CRDT, deferred per Founder decision #2)
- **Stable identity**: new `uuid` column (fresh **UUIDv7** assigned on first sync — Founder decision #3, not content-derived UUIDv5) plus a `remote_id` column mapping local rows to the cloud-minted id. Partial-unique indexes; nullable so legacy rows migrate lazily.
- Per-project `sync_state` watermark table (ISO 8601 cursors; no PII).
- `spelunk memory pull`: delta pull from `GET /memory/since?t=<watermark>`, persists `last_synced`.
- `spelunk sync` (and the top-level `sync` alias): batched push via `POST /memory/batch` (not N single POSTs) + delta pull, applied **keep-both** (ADR-005 Add-Wins). **Idempotent** — identity-keyed dedupe means re-running never duplicates (server skips by `external_id`; pull dedupes by `remote_id`).
- **Tombstone propagation**: a local archive propagates to the cloud via `DELETE /memory/{id}`; a pulled tombstone archives the local copy (never un-archives).

### D3 — embedding conformance
- `push`/`sync` send **text-only**; the old client-vector path (`local.get_embedding(...)`) is removed. No `--client-embed` escape hatch (ADR-020).

## OSS changelog / migration impact
**New local schema columns (`notes.uuid`, `notes.remote_id`) + `sync_state` table.** The migration is additive and idempotent (guards duplicate-column on re-open); existing rows get a UUID lazily on first sync. **Breaking for scripts relying on client-vector push**: `memory push` is now text-only and idempotent (the server backfills embeddings). Please call this out in the OSS changelog.

## Test / quality status
- `cargo fmt --check`: clean (enforced by pre-commit hook).
- `cargo clippy -p spelunk-core -p spelunk-cli --all-targets -- -D warnings`: clean.
- `cargo test`: green (core + cli + server build). New tests: SyncMode parse/defaults, backend-selection per mode (+ kill-switch), UUID backfill idempotency, watermark round-trip, idempotent apply (no dupes), tombstone reconcile, wiremock coverage for text-only push / delta pull / DELETE-404-idempotent.
- `cargo audit`: unchanged — no new dependencies (`uuid` v7 + `chrono` were already in the workspace). Only the pre-existing, allowlisted `paste` unmaintained advisory remains.

## Spec ambiguities resolved (please confirm with Architect/Founder)
1. **`local_first` routes memory CRUD to the *local* store**, per the ADR-037 D1 table (reads local; writes local then async-sync). This is a **behaviour change for teams with an explicit `server_url`**: pre-ADR-037, an explicit `server_url` relocated memory to the shared server (ADR-004 team-memory tier); now the default `local_first` keeps it local and converges via `spelunk sync`. Cloud-routed CRUD is reachable via `cloud_first` (debug/override). Flagging because it touches the ADR-004 team-memory model.
2. **`project_id` for cloud-api must be the project UUID.** cloud-api's memory routes are typed `Path<Uuid>`, so dogfooding against `api.spelunk.cloud` requires `project_id = "<uuid>"` in config (the existing slug-based `RemoteMemoryBackend` only matches the OSS spelunk-server). No new resolution layer was added — out of P1 scope.
3. **`supersedes` edge cannot be pushed through existing endpoints.** cloud-api's batch/create payloads expose no `supersedes` field and there is no CLI-reachable supersede-edge endpoint. The *effect* of supersede still propagates: superseding locally archives the predecessor, whose tombstone propagates via DELETE on `--include-archived`. Explicit edge propagation would need a new cloud endpoint (out of P1 scope — flagging for the Architect).

## Deferred (not in this PR, per scope)
P2 background reconciler / write-queue / SSE live-pull; P3 `spelunk login` device flow. Cloud-api endpoints were wired to, not built.


crates/spelunk-core/migrations/021_sync_state.sql
```diff
@@ -0,0 +1,20 @@
...
+--
+-- `spelunk sync` / `memory pull` persist the high-water mark of the last
+-- successful pull (and the last push) so subsequent runs only transfer the
+-- delta. Keyed by the server-side project identifier so a single local store
+-- that syncs against multiple projects keeps independent cursors.
+--
+-- Timestamps are RFC 3339 / ISO 8601 strings to match cloud-api's
+-- `created_at` wire format (the `since` endpoint takes and returns ISO 8601),
+-- avoiding lossy epoch round-tripping.
+--
```
usercise: This sounds a a frail solution. What happens if there is drift between the local and remote time? Since we're now using uuid v7 and that starts with a timestamp could we not just rely on the last remote uuid that was synced? Something like `sync to server WHERE remote_uuid = NULL` & `sync form server WHERE uuid > last_remote_uuid`.

usercise: Good catch — switched to a UUIDv7 identity cursor exactly as you suggested (decision #183). Push set is now local rows `WHERE remote_id IS NULL`; pull asks the server for `uuid > last_remote_uuid` (= `MAX(remote_id)` of already-synced rows). The timestamp watermark (`021_sync_state.sql`) is dropped, so there is no clock-drift dependency. Boundary same-millisecond rows are harmless because pull dedupes by `remote_id` (idempotent).

The server side shipped in cloud-api **PR #96 (merged)** — `/v1/projects/{id}/memory/since` now accepts a `since_id` UUID cursor (the `t` timestamp param stays for SSE catch-up). This commit (`22eb531d`) wires the CLI to it.

PR #453, opened 2026-06-28
release: Windows artifacts (.zip) + install.ps1 + docs

usercise:
## Summary

- Adds `x86_64-pc-windows-msvc` to the `release.yml` build matrix (`windows-latest` runner). The build step uses `shell: bash` for cross-platform compatibility (Git Bash on Windows). A Windows-specific packaging step produces `spelunk-<version>-x86_64-pc-windows-msvc.zip` containing `spelunk.exe` and `spelunk-server.exe` via `Compress-Archive`. The zip is attached to the GitHub release alongside the existing `.tar.gz` and `.deb` assets.
- Adds `install.ps1` at the repo root: mirrors `install.sh` in structure and UX — detects arch, resolves the latest release tag via the GitHub API, downloads the `.zip`, extracts both binaries to `%LOCALAPPDATA%\Programs\spelunk\`, and adds that directory to the user PATH (registry) automatically. Supports `-DryRun`. Requires PowerShell 5.1+.
- Updates `docs/getting-started.md`: adds a Windows install section (PowerShell script + manual zip) above the macOS/Linux section; replaces the stale `universal-apple-darwin` manual-tarball example with a per-platform table listing all four shipped targets; removes the stale `x86_64-apple-darwin` target reference.
- Updates `docs/releasing.md`: adds Windows row to the supported platforms table with archive format column; adds Windows zip to the download URL examples.

## Deferred follow-ups (not in this PR)

**winget manifest / Scoop bucket** — Johan's preference is winget as the primary managed Windows install path. These are deferred pending a founder/GTM call on which managed package channel(s) to support and the associated submission process. The `install.ps1` script is the in-repo default in the meantime.

**Windows ARM64** — `aarch64-pc-windows-msvc` is not added; `install.ps1` exits with a clear message on ARM64 machines pointing to build-from-source.

## Test plan

**Validated locally:**
- YAML is well-formed (no tab characters; all structural elements confirmed present via grep)
- Windows build path traced end-to-end: matrix entry → `shell: bash` build step → `shell: pwsh` zip step → `GITHUB_ENV` write → artifact upload → release job `*.zip` glob → attached to release
- Homebrew update job is untouched: reads only `.tar.gz` digests for unix targets
- `install.ps1`: arch detection enum values correct for `RuntimeInformation.ProcessArchitecture`; error handling with `$ErrorActionPreference = 'Stop'`; temp dir cleanup in `finally`; user PATH update via registry; `-DryRun` switch works before any network call
- `rich-formats` feature dependencies (`lopdf`, `docx-rs`, `calamine`) are pure Rust with no Unix-only deps
- `cargo fmt --check` + `cargo clippy` both passed (pre-commit hooks on both commits)
- No internal workstream names, ADR references, or cloud-product details in any changed file (OSS boundary check passed)

**Can only be verified by an actual tagged release run in CI:**
- Windows build succeeds on `windows-latest` with MSVC toolchain and `rich-formats` feature
- `Compress-Archive` produces a correctly-structured zip on the runner
- `gh release create ... artifacts/*.zip` glob picks up the zip and attaches it to the release
- `install.ps1` end-to-end: download, extract, PATH update — on a real Windows machine

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -9,11 +9,51 @@ at the end).
 
 ## 1. Install spelunk
 
+### Windows
+
+#### Install script (PowerShell) — recommended
+
+The PowerShell install script resolves the latest release, downloads the
+Windows `.zip`, and installs `spelunk.exe` and `spelunk-server.exe` to
+`%LOCALAPPDATA%\Programs\spelunk\`. It also adds that directory to your user
+`PATH` automatically.
+
+Open PowerShell and run:
+
+```powershell
+irm https://spelunk.cloud/install.ps1 | iex
+spelunk --version
+```
+
+Preview what it would do without writing anything:
+
+```powershell
+& ([scriptblock]::Create((irm https://spelunk.cloud/install.ps1))) --dry-run
+```
+
+#### Manual `.zip` download
+
+Download the `.zip` for your platform from the
+[releases page](https://github.com/spelunk-cloud/spelunk/releases). The Windows
+archive is named `spelunk-<version>-x86_64-pc-windows-msvc.zip`. Extract it and
+place `spelunk.exe` and `spelunk-server.exe` anywhere on your `PATH`
+(e.g. `C:\Users\<you>\bin\`).
+
+> **Windows package managers (winget, Scoop):** a winget manifest is planned as
+> the primary managed install path for Windows; submission is pending a decision
+> on which managed channel to support. Follow the
+> [releases page](https://github.com/spelunk-cloud/spelunk/releases) or the
+> repo for updates.
+
+---
+
+### macOS and Linux
+
 The recommended install paths are Homebrew (macOS/Linux), the install script,
 and the Debian package (Linux). All three drop both `spelunk` and
 `spelunk-server` onto your `$PATH`.
 
-### Install script (macOS and Linux) — recommended
+#### Install script (macOS and Linux) — recommended
 
 Detects your OS/arch, resolves the latest release tag via the GitHub API,
 downloads the matching tarball, and installs both binaries to `/usr/local/bin`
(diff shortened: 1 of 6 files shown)


docs/getting-started.md
```diff
@@ -9,11 +9,51 @@ at the end).
...
+
+The PowerShell install script resolves the latest release, downloads the
+Windows `.zip`, and installs `spelunk.exe` and `spelunk-server.exe` to
+`%LOCALAPPDATA%\Programs\spelunk\`. It also adds that directory to your user
+`PATH` automatically.
+
+Open PowerShell and run:
+
+```powershell
+irm https://spelunk.cloud/install.ps1 | iex
```
usercise: use raw githubcontent instead of spelunk.cloud so the install script doesn't get out of sync.

usercise: Done in `029217ab` — the Windows one-liner now uses `https://raw.githubusercontent.com/spelunk-cloud/spelunk/main/install.ps1` (in both `docs/getting-started.md` and the `install.ps1` usage header), so the documented command always matches the committed script. Left the unix `install.sh` on its existing `spelunk.cloud/install.sh` URL since that redirect is already set up and documented in `releasing.md`; happy to switch that too if you want symmetry.

usercise: Correction to my previous reply — you were right, I had it backwards. `spelunk.cloud/install.sh` is **broken**, not canonical (Vercel does not serve `.sh` from `public/`, per marketing-site CLAUDE.md), and the marketing docs are the source of truth using the GitHub raw URL. Fixed **all** install one-liners — `README.md`, `docs/getting-started.md`, `docs/releasing.md`, and the `install.sh` / `install.ps1` usage headers — to `https://raw.githubusercontent.com/spelunk-cloud/spelunk/refs/heads/main/install.{sh,ps1}`, matching the marketing site exactly. No `spelunk.cloud/install.*` references remain (only the homepage branding line in each script header). Pushed in `e8bcf27c`.

PR #500, opened 2026-07-02
security(server): ADR-056 single-trust-domain guardrails (warning + docs)

usercise:
## Summary

Per [ADR-056](docs/adr/056-oss-server-tenancy-model.md) (ratified, merged), a `spelunk-server` instance is a single trust domain: the shared API key is the tenancy boundary, and every keyholder is a full administrator of every project on the instance. Cross-project access via a shared key is intended behaviour, not a defect — but it must be explicit and impossible to stumble into. This PR ships the guardrails ADR-056 calls for.

- **Startup warning:** `spelunk-server` now emits a prominent `tracing::warn!` when it binds a non-loopback address with a key configured (a shared/team deployment), restating that every keyholder can read, modify and permanently delete all projects' memory on the instance and that separate server instances (separate keys) are the way to get isolation. Suppressed on loopback binds and when no key is set.
- **Docs:** adds a "Trust model" section to `docs/server.md`, a pointer from `docs/self-hosting.md`, and a callout in `examples/mdm/README.md`'s shared-server deployment shape — all explaining the single-trust-domain model and linking ADR-056.
- **API docs:** `GET /v1/projects` now documents (utoipa doc comment, regenerated into `docs/openapi.json`) that it enumerates every project on the instance by design.

No handler behavior changes and no new ACL — that's the ratified decision. `GET /v1/health` was already outside the auth middleware (unauthenticated by construction) and needed no change.

## Test plan

- [x] `SPELUNK_SECRET_STORE=file cargo test -p spelunk-server` — all 21 unit tests + 12 integration tests pass, including 3 new tests on the extracted warning predicate:
  - `trust_domain_warning_fires_for_non_loopback_with_key` — fires for `0.0.0.0`, `::`, a LAN IP, and a hostname, all with a key set.
  - `trust_domain_warning_suppressed_on_loopback` — suppressed for `127.0.0.1`/`::1`/`localhost` regardless of key.
  - `trust_domain_warning_suppressed_without_key` — suppressed when no key is configured.
- [x] `cargo fmt --all -- --check` clean.
- [x] `cargo clippy --workspace --all-targets -- -D warnings` clean.
- [x] Manual: `SPELUNK_SECRET_STORE=file cargo run -p spelunk-server -- --host 0.0.0.0 --key test` and confirmed the warning line appears in the startup log; `--host 127.0.0.1 --key test` and confirmed it does not.

### DIFF
--- a/crates/spelunk-server/src/handlers.rs
+++ b/crates/spelunk-server/src/handlers.rs
@@ -225,6 +225,10 @@ fn require_embedder(
 // ── Projects ──────────────────────────────────────────────────────────────────
 
 /// List all projects registered on this server.
+///
+/// Enumerates every project on the instance, by design: this server is a
+/// single trust domain (ADR-056) and any valid key is a full administrator of
+/// every project on it, so there is no per-caller filtering to apply here.
 #[utoipa::path(
     get,
     path = "/v1/projects",
--- a/crates/spelunk-server/src/main.rs
+++ b/crates/spelunk-server/src/main.rs
@@ -126,6 +126,16 @@ async fn main() -> Result<()> {
         );
     }
 
+    // Single-trust-domain notice (ADR-056): a keyed, non-loopback bind is a
+    // shared/team server. The shared key is the *only* boundary — every
+    // keyholder is a full administrator of every project on this instance
+    // (list, read, write, supersede, archive, delete). This is intended
+    // behaviour, not a bug; teams that must not see each other's memory need
+    // separate server instances (separate keys, separate databases), not a
+    // per-project ACL on one instance. Loopback binds are a single developer's
+    // own machine, so the notice does not apply there.
+    warn_single_trust_domain(&args.host, api_key.is_some());
+
     // Build the auth provider from the configured key.
     let auth: Arc<dyn spelunk_server::auth::AuthProvider> =
         Arc::new(ApiKeyAuth::new(api_key.clone()));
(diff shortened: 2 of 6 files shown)


usercise (commented):
## QA review — approve

Reviewed against ADR-056 (merged, PR #499) and the task's engineer-scope acceptance criteria. Verified independently (not just from the PR description): pulled the branch, read the diff line-by-line, and ran the suite myself.

**Predicate logic (`should_warn_single_trust_domain`, `main.rs`):** `!host_is_loopback(host) && key_is_set`. Traced through `host_is_loopback` (handles `localhost`, and any IP that parses `.is_loopback()`, so `127.0.0.1`/`::1` both covered). This is exactly the fire/suppress matrix the task asked for:
- Fires: non-loopback + key set (`0.0.0.0`, `::`, LAN IP, hostname).
- Suppressed: loopback (with or without a key), and no-key case (belt-and-suspenders, since `check_bind_safety` already refuses a keyless non-loopback bind before this runs).

**Tests:** ran `SPELUNK_SECRET_STORE=file cargo test -p spelunk-server` myself — 35+21+12 passed, 0 failed, including the 3 new predicate tests (`trust_domain_warning_fires_for_non_loopback_with_key`, `_suppressed_on_loopback`, `_suppressed_without_key`) and the pre-existing `health_returns_ok` / `protected_endpoint_rejects_missing_token` integration tests, which stay green — confirming no ACL/handler regression. `cargo fmt --all -- --check` and `cargo clippy -p spelunk-server --all-targets -- -D warnings` both clean on my run too. Also confirmed `default_host_is_loopback` test asserts the binary default is `127.0.0.1`, per the task's flagged concern about a possible `0.0.0.0` default.

**Docs:** `docs/server.md` "Trust model" section, `docs/self-hosting.md` pointer, and `examples/mdm/README.md` shared-server callout all state the single-trust-domain model plainly and link `docs/adr/056-oss-server-tenancy-model.md`. `GET /v1/projects` gets both a utoipa doc comment and a `docs/server.md` note that enumerate-all is by design, and `docs/openapi.json` was regenerated to match (verified the diff contains the added `description` field).

**No ACL/handler behavior change:** confirmed — the diff touches only `main.rs` (new warn fn + tests), a doc-comment-only addition to `handlers.rs`'s `list_projects`, `docs/openapi.json`, and three markdown files. `auth.rs` and `db.rs` are untouched, so the cross-project access described in ADR-056 is unchanged, as scoped.

**`/v1/health`:** spot-checked `lib.rs` router assembly — `/v1/health` is registered on the outer `Router::new()` *before* `.merge(protected)`, and `auth_middleware` is layered only onto the `protected` sub-router. So it's genuinely unauthenticated by construction; the engineer's claim that no code change was needed here is correct, not just asserted.

**On the skipped "hard refusal of a keyed non-loopback plaintext bind" item:** sanity-checked this rather than taking it on faith. ADR-056's own "Guardrails required" section (the ratified spec this PR implements) says only: *"The server emits a prominent startup warning when it is bound to a non-loopback interface and a key is set... "* — a warning, not a refusal. Separately, the shipped `Dockerfile` CMD is `["--host", "0.0.0.0", ...]` with no TLS termination in the container, and `docker-compose.yml`'s documented quick-start explicitly supports running with an empty `SPELUNK_SERVER_KEY` (`${SPELUNK_SERVER_KEY:-}`) — i.e. the primary Docker path *is* a plaintext non-loopback bind, sometimes keyed. A hard refusal as literally worded in the task's later-added acceptance criteria would break that shipped default without a corresponding Dockerfile/compose/proxy rework. The engineer's call to treat that as scope creep beyond the merged ADR and flag it for separate, coordinated work (with oss^63 client-side validation and oss^66 audit reconciliation) rather than silently implementing a breaking change is the right call. This is a real gap between the merged ADR text and the task description's later v4 edit ("Johan's transport requirement") that's worth the architect/EM reconciling explicitly — flagging on the board, not blocking this PR for it.

**Verdict: approve.** Guardrail warning logic is correct and tested both ways, docs are clear and correctly cross-linked, no scope creep into ACL/handler behavior, and the one deliberately-skipped item is a defensible, well-reasoned scope call backed by the actual shipped Docker config — not a rubber-stamped excuse.

PR #515, opened 2026-07-04
refactor(embed): extract native F2LLM embedder into spelunk-embed crate with local-path load (oss^82)

usercise:
## What

Extracts the candle F2LLM-v2-330M native embedder out of the `spelunk-server` binary crate into a new workspace **library** crate, `crates/spelunk-embed`, and adds a zero-network local-path loader.

- **New crate `spelunk-embed`** (lib): owns `embedder_native.rs` (the Qwen3 decoder, 896-dim, Q8_0 GGUF engine). The heavy candle stack (`candle-core`/`candle-nn`/`candle-transformers`, `hf-hub`, `tokenizers`) stays behind the crate's `embed-native` feature (on by default); `metal` forwards Metal GPU acceleration on macOS.
- **Public API:**
  - `NativeEmbedder::load_from_hub()` (the former `load()`, renamed): downloads/quantizes via the Hugging Face Hub cache. Behaviour unchanged.
  - `NativeEmbedder::load_from_path(gguf_path: &Path, tokenizer_path: &Path, config_path: &Path) -> anyhow::Result<Self>` (new): loads the model from local files already on disk with no network access. Fails fast if the GGUF is absent, then reads the tokenizer/config from the given paths and builds the same `NativeEmbedder`. No `hf_hub` API is constructed on this path.
  - `DIM` (896) and `NativeEmbedder` are re-exported from the crate root, gated by `embed-native`.
  - `embed()` behaviour is byte-identical (896-dim, L2-normalised, same query/document prompt formatting).
- **`spelunk-server` rewired** to a workspace path-dep on `spelunk-embed` (`default-features = false, optional = true`); its `embed-native` / `metal` features forward to the crate. The `use spelunk_embed::...` import is `#[cfg(feature = "embed-native")]`-gated so a no-embed build still compiles. No functional change to the server or its embed endpoints.
- Docs: `docs/architecture.md` and `docs/third-party-models.md` (the `MODEL_REVISION` pin citation) now point at the new crate path; `CLAUDE.md` module map updated. Pure internal refactor, so no CHANGELOG entry is owed (zero user-facing behaviour change).

This is the OSS half. It unblocks **cloud-api^78**, which will git-dep this crate, fetch the digest-pinned model artifact from GCS to a temp file, and call `load_from_path`.

## Test plan

All commands run with `SPELUNK_SECRET_STORE=file`.

- `cargo build --workspace` -> green. `spelunk-embed` compiles standalone; `spelunk-server` builds against it via the path-dep.
- `cargo check -p spelunk-server --no-default-features` (forced fresh, after touching `main.rs`/`lib.rs`) -> green. Confirms the cfg-gated `spelunk_embed` import drops cleanly and a no-`embed-native` server build still compiles.
- `cargo test --workspace` -> green, 0 failed. Notable: `spelunk-embed` = 11 passed + 4 ignored (the ignored 4 are model-download e2e cases). `spelunk-server` lib 65, bin 12, `integration_server` 12; `spelunk-core` + `spelunk-cli` suites all green.
- No-network guarantee for `load_from_path`, covered by two passing unit tests:
  - `load_from_path_missing_gguf_errors_without_network`: a missing GGUF fails fast on the local `exists()` check.
  - `load_from_path_present_gguf_missing_tokenizer_errors_locally`: with a present-but-empty GGUF the existence guard passes, so control reaches the tokenizer read; the test asserts the error names the on-disk tokenizer path and contains no `http` / `huggingface` / `downloading`. This is the regression guard against a future Hub fallback slipping into the post-existence-check branch.
- `cargo fmt --check` -> clean.
- `cargo clippy --workspace -- -D warnings` -> clean; plus a fresh `cargo clippy -p spelunk-embed --all-targets --all-features -- -D warnings` -> clean.
- Staleness: fetched `origin/main`; the only commit on `main` since this branch's base is #514 (oss^80), which touches `crates/spelunk-cli/src/cli/cmd/server.rs` only. This branch does not touch that file; a merge dry-run of `origin/main` into the branch is conflict-free with no revert.

Do not merge. Held for founder review.


Cargo.toml
```diff
@@ -23,6 +24,18 @@ tempfile    = "3"
 pretty_assertions = "1"
 serial_test = "3"
 wiremock    = "0.6"
+dirs        = "6"
+# `libc` is macOS-only here: the native embedder reads `hw.memsize` via the
+# `sysctlbyname` FFI to size its memory budget (Linux reads /proc/meminfo).
+libc        = "0.2"
```
usercise: This was `[target.'cfg(target_os = "macos")'.dependencies]` gated before.

crates/spelunk-embed/Cargo.toml
```diff
@@ -0,0 +1,43 @@
...
+# macOS-only: `hw.memsize` sysctl in `total_system_ram` (Linux reads
+# /proc/meminfo, no FFI).
+[target.'cfg(target_os = "macos")'.dependencies]
+libc = { workspace = true }
+
+[dev-dependencies]
+tempfile = { workspace = true }
+pretty_assertions = { workspace = true }
+serial_test = { workspace = true }
+tokio = { workspace = true }
```
usercise: Why do we need this both as dependency and dev dependency?

crates/spelunk-embed/src/embedder_native.rs
```diff
@@ -546,7 +546,10 @@ impl NativeEmbedder {
     /// access. The tokenizer and config are always fetched from the pinned
     /// upstream revision. Uses Metal/GPU on macOS when built with the `metal`
     /// cargo feature, CPU otherwise.
-    pub fn load() -> Result<Self> {
+    ///
+    /// For a fully offline load from files already on disk, see
+    /// [`NativeEmbedder::load_from_path`].
+    pub fn load_from_hub() -> Result<Self> {
```
usercise: This should live in server crate not embed. We don't want to carry the hf crate dependency into cloud. This crate should always be called through load_from_path.

Cargo.toml
usercise: Restored (0ec41d92): the extraction had silently promoted libc to an ungated [workspace.dependencies] entry. It's now pinned inside spelunk-embed's [target.'cfg(target_os = "macos")'.dependencies] table — its sole consumer — matching the pre-extraction shape. Verified per-target: direct dep on aarch64-apple-darwin, absent on x86_64-unknown-linux-gnu (only candle's transitive copy remains there).

crates/spelunk-embed/Cargo.toml
usercise: Fully redundant — the crate has no separate tests/ dir and the [dependencies] tokio (workspace features) already covers the inline #[cfg(test)] module. Dev-dep entry removed (0ec41d92).

crates/spelunk-embed/src/embedder_native.rs
usercise: Moved (0ec41d92): the whole hub-acquisition path (load_from_hub + download/quantize/cache helpers and their constants/tests) now lives in a new feature-gated crates/spelunk-server/src/embed_hub.rs; spelunk-embed's public API is load_from_path + DIM only. Evidence: `cargo tree -p spelunk-embed -i hf-hub` no longer matches any package; hf-hub/dirs are optional deps of spelunk-server behind embed-native; `cargo check -p spelunk-server --no-default-features` clean. Docs (architecture, third-party-models, CLAUDE.md) repointed.

crates/spelunk-server/src/embed_hub.rs
```diff
@@ -0,0 +1,522 @@
...
+///   (`f2llm-v2-330m-q8_0.gguf`) straight from `spelunk-cloud/F2LLM-v2-330M-Q8_0-GGUF`
+///   through the hf-hub cache (checksum/resume reused) — first-run download is
+///   ~339 MB and there is no on-device quantize step. Set
+///   `SPELUNK_EMBEDDER_GGUF_REPO` to a different `org/repo` to fetch the
+///   pre-quant GGUF from there instead.
+/// * **On-device quantize (escape hatch, `SPELUNK_EMBEDDER_GGUF_REPO=off`):**
+///   download the ~638 MB BF16 safetensors from the pinned upstream revision
+///   of `codefuse-ai/F2LLM-v2-330M`, quantize on device to a ~339 MB GGUF,
+///   then delete the cached safetensors so steady-state disk is ~339 MB
+///   rather than ~1.5 GB.
```
usercise: We can delete this now that we have the new delivery in place and tested. A fall back for a first party file not existing that rely on a third party not to delete their files isn't really a reliable fallback.

crates/spelunk-server/src/embed_hub.rs
```diff
@@ -0,0 +1,522 @@
...
+            })
+            .collect();
+    }
+    Ok(vec![
+        repo.get("model.safetensors")
+            .context("downloading model.safetensors")?,
+    ])
+}
+
+/// GGML dtype to quantize a given (prefixed) weight key to, or `None` to skip it.
```
usercise: No need for local quantize in v1.

Cargo.toml
```diff
@@ -23,6 +24,15 @@ tempfile    = "3"
...
 serial_test = "3"
 wiremock    = "0.6"
+dirs        = "6"
+# Native embedder stack (F2LLM-v2-330M via candle). Pulled in by the
+# `spelunk-embed` crate behind its `embed-native` feature; versions pinned here
+# so the candle trio stays on one runtime line.
+candle-core = "0.11.0"
+candle-nn   = "0.11.0"
+candle-transformers = "0.11.0"
+hf-hub      = { version = "0.5", default-features = false, features = ["tokio", "ureq", "rustls-tls"] }
```
usercise: This is only used in server crate so can safely move back there, can it not?

PR #517, opened 2026-07-04
feat!: remove non-functional search --as-of snapshot feature (spelunk-oss^67)

usercise:
## Summary

Removes `spelunk search --as-of <sha>` (snapshot-based temporal search) and the
dead snapshot storage layer behind it, per a founder-approved decision to drop
the feature outright rather than fix or cfg-gate it.

## Why

The snapshot storage layer was never wired to the indexer: nothing in the
indexing path ever called `create_snapshot` / `insert_snapshot_*` /
`update_snapshot_stats`, so `list_snapshots()` was always empty and the
`--as-of <sha>` flag errored on every single use. It shipped non-functional.
Temporal/as-of semantic search over code snapshots is deferred to a future
design rather than patched in place.

## What's removed

- The `--as-of <sha>` flag and its handling from `spelunk search`.
- The snapshot storage layer (`create_snapshot`, `insert_snapshot_*`,
  `update_snapshot_stats`, `list_snapshots`, and related plumbing).
- No `cfg`-gating — this is a full removal, not a feature flag.

## Migration note

Migration `021_drop_snapshots.sql` drops the four dead snapshot tables
(`snapshots`, `snapshot_files`, `snapshot_chunks`, `snapshot_embeddings`) on
any database opened with this version. `snapshot_embeddings` is a
`sqlite-vec` `vec0` virtual table and is dropped defensively before its
logical parents. This runs automatically on next open; no user action
required.

## Unaffected: `spelunk memory list/search --as-of <date>`

This is a **different, live feature** — point-in-time memory archaeology over
the memory store — and is untouched by this change. A regression test
(`test_search_help_does_not_list_as_of` in
`crates/spelunk-cli/tests/e2e_cli.rs`) asserts `spelunk search --help` no
longer lists `--as-of` while `spelunk memory search --help` still does, so the
two flags can't be confused going forward.

## Test plan

All commands run from a clean worktree with `SPELUNK_SECRET_STORE=file` set
(avoids macOS Keychain prompts).

- [x] `cargo test --workspace` — full workspace suite green:
  all test binaries reported `test result: ok. ... 0 failed`
  (e.g. spelunk-core lib: 179 passed; spelunk-server: 214 passed;
  spelunk-cli e2e/integration suites: all passed; 0 failures anywhere).
- [x] `cargo test -p spelunk-core --test drop_snapshots_migration fresh_database_has_no_snapshot_tables -- --exact`
  — passes in isolation (`test result: ok. 1 passed; 0 failed`), confirming the
  test-isolation bug from round 1 (missing `common::register_sqlite_vec()`
  call) is fixed and the test doesn't depend on ordering/leakage from other
  tests in the binary.
- [x] `drop_snapshots_migration_removes_tables_with_preexisting_rows` (same
  test file) — verifies migration 021 drops all four snapshot tables from a
  pre-^67 database that was seeded with real snapshot rows.
- [x] `test_search_help_does_not_list_as_of` (`crates/spelunk-cli/tests/e2e_cli.rs`)
  — help-surface guard: `spelunk search --help` no longer lists `--as-of`,
  while `spelunk memory search --help` still does (proves the unrelated
  `memory --as-of <date>` feature is untouched).
- [x] `cargo fmt --check` and `cargo clippy` — clean (also enforced via
  pre-commit hook on every commit in this branch).
- [x] `git fetch origin` + `git merge-base --is-ancestor origin/main HEAD` —
  confirms this branch contains all of `origin/main` with no drift.


crates/spelunk-core/tests/drop_snapshots_migration.rs
usercise, line 1: I don't see the value of this test. It will never run in production which is where we care about it the most. The fact CI passed on the branch is as much value as we will ever get out of it, but we will carry the slowdown in CI build indefinitely if we keep it. Let's delete the file.

usercise: Deleted (e241be966). Removed `crates/spelunk-core/tests/drop_snapshots_migration.rs` entirely — the CI slowdown isn't worth a one-time migration proof that never runs in production. The cheap CLI help-surface guard (`test_search_help_does_not_list_as_of`) stays, since it guards the actual user-facing removal on every build. Workspace green.

PR #516, opened 2026-07-04
security(server): refuse keyed non-loopback plaintext bind by default (oss^79)

usercise:
## Summary

Tightens the `spelunk-server` startup bind guard to match the ADR-056 transport guardrail ("loopback-only plaintext; TLS-front for non-loopback").

Previously `check_bind_safety` only refused a **keyless** non-loopback bind (an open, unauthenticated server). A **keyed** non-loopback bind over plaintext HTTP was still allowed, so a shared server on a routable interface sent the bearer `SPELUNK_SERVER_KEY` across the network in cleartext. This PR refuses a non-loopback plaintext bind by default whether or not a key is set. The refusal names the interface/port and points at the loopback-only / TLS-front guidance and the ADR.

A new opt-out, `--allow-insecure-remote` / `SPELUNK_ALLOW_INSECURE_REMOTE` (accepts `1`/`true`/`yes`/`on`; set-but-empty counts as unset), downgrades **only** the keyed-plaintext refusal to a loud one-time startup `tracing::warn!` and proceeds. It never lifts the keyless refusal. This is the sanctioned mechanism for a private, trusted network and for the Docker remote-agent workflow (`host.docker.internal` / `172.17.0.1`, spelunk-oss^70), documented as such in `server.md` and `remote-agents.md`.

Loopback binds (the default) and TLS-fronted loopback deployments are unaffected.

Closes the doc-vs-code caveat from spelunk-oss^66 (V1-SERVER-AUDIT §3/§7 now describe the as-shipped refuse-by-default + named opt-out) and provides the sanctioned insecure-remote mechanism for spelunk-oss^70.

Refs: ADR-056 (PR #512), spelunk-oss^66, spelunk-oss^70, spelunk-oss^79.

## Changes

- `crates/spelunk-server/src/main.rs`: `check_bind_safety(host, key_is_set, allow_insecure)` refuses a keyed non-loopback plaintext bind by default; new `--allow-insecure-remote` flag + `parse_env_flag` helper.
- `docs/adr/056-oss-server-tenancy-model.md`: Consequences + Security-implications describe refuse-by-default (keyless + keyed) plus the named accepted-risk opt-out.
- `docs/security/V1-SERVER-AUDIT.md`: §3 new transport-guardrail paragraph; §7 JWT row relabelled to as-shipped.
- `docs/server.md`: bind/auth flags table + "Insecure non-loopback plaintext bind opt-out" section with the cleartext-key caveat and a `172.17.0.1` example.
[...]

PR #521, opened 2026-07-05
docs(adr): ADR-058 bare-metal + systemd as recommended team-server deployment

usercise:
## What

Records the decision (Johan, 2026-07-05, on PR #516) that **bare-metal/systemd is the recommended `spelunk-server` team-server deployment**, with Docker Compose demoted to a minimal local scaffold. Follows from PR #516's unconditional refusal of non-loopback plaintext binds: a container's loopback bind is unreachable off-container, so containers are a poor fit for the loopback-by-design server, whereas same-host bare-metal (server on 127.0.0.1 + an operator-owned TLS terminator on the same host) works.

## What it decides

- **Topology:** server binds `127.0.0.1` + key; operator-owned TLS terminator on the same host forwards to loopback; clients point `server_url` at the operator's `https://` hostname.
- **Packaging:** ship a first-party `spelunk-server.service` systemd unit (static `spelunk` user, `LoadCredential` for the key, `ProtectSystem=strict`/`NoNewPrivileges`). Do **not** ship any proxy config — reference examples only, clearly operator-owned.
- **Key provisioning:** systemd `LoadCredential` preferred over `Environment=`.
- **Docs restructure** captured as a plan (not yet executed): elevate bare-metal in `server.md`/`self-hosting.md`, keep Docker local-only, supersede PR #516's interim pointer. Folds in the Docker remote-agent friction (containerize the agent, not the server).
- **Non-goal:** native TLS in the server (a separate, larger future decision).

## Open questions for review (recorded in the ADR, not decided)

1. **Credential wiring:** an `ExecStartPre` wrapper exporting the env var (no code change) vs teaching the binary a `--key-file`/`$CREDENTIALS_DIRECTORY` read (small code addition). ADR leans to the latter.
2. Whether to also ship a `DynamicUser=` variant alongside the static-user default.

## Review note

This is a decision record for review — **no implementation is dispatched until it's approved/merged** (ADR-approval gate). Next free ADR number verified globally across both repos (max was 057).

### DIFF
--- a/docs/adr/058-team-server-bare-metal-deployment.md
+++ b/docs/adr/058-team-server-bare-metal-deployment.md
@@ -1,0 +1,344 @@
+# ADR-058: Recommended team-server deployment is bare-metal + systemd; Docker is a local scaffold
+
+**Date:** 2026-07-05
+**Deciders:** founder (Johan), architect
+**Trigger:** founder review on PR #516 (spelunk-oss^79). The pre-v1.0 server
+hardening made `spelunk-server` refuse a non-loopback plaintext bind
+*unconditionally*, and the founder review of that PR removed the bundled
+reverse-proxy/TLS-sidecar packaging entirely (Caddyfile and
+`docker-compose.full.yml` deleted; the proxy remediation text stripped from the
+refusal message, ADR-056, and `server.md`). That left `server.md` describing a
+Docker quick-start that publishes a port nothing can reach, and no first-class
+answer to "how do I actually stand up a shared team server?" This ADR records
+the founder's answer.
+
+---
+
+## Context
+
+`spelunk-server` (`crates/spelunk-server/`) is an axum HTTP listener. It serves
+two roles ([`server.md`](../server.md)): an automatic loopback **inference**
+backend on a developer's own machine (never a memory store), and — when a
+developer sets an explicit `server_url` — a deployed **team memory** server that
+holds a team's shared memory ([ADR-004](004-unified-memory-storage.md),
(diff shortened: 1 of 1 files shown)


docs/adr/058-team-server-bare-metal-deployment.md
```diff
@@ -0,0 +1,344 @@
...
+  loopback.
+
+Only a **same-host bare-metal** process can bind the host's real loopback and be
+reached there by a same-host TLS terminator. This is the mechanical reason
+bare-metal — not Docker — is the deployment vehicle for a networked team server.
+
+This also resolves the [spelunk-oss^70](../remote-agents.md) friction (Docker
+remote-agent loopback): the fix is to containerize the **agent**, not the
+**server**. With the server on bare-metal loopback, a containerized agent
+reaches it over the Docker bridge / `host.docker.internal`
```
usercise: Flagging for tightening during review (raised by the EM re: whether this resolves spelunk-oss^70): this sentence overclaims for the native-Linux case, which is the bare-metal host this ADR recommends. On native Linux, `host.docker.internal`/`172.17.0.1` resolves to the docker0 bridge gateway — a server bound to the host's `127.0.0.1` is not listening there, so the containerized-agent → bridge → loopback path does NOT work (it only works on Docker Desktop, or with `--network host` on the agent container). The path that actually works under this ADR's model is: agent → the operator's `https://` TLS hostname → operator proxy → `127.0.0.1` server. Suggest rewording so the remote agent is a client of the TLS endpoint, not the raw bridge/loopback — and making the remote-agents.md R1 recipe rework (the concrete oss^70 fix) an explicit item in the §4 docs restructure.

PR #531, opened 2026-07-06
Freeze git-notes v1 format (JSONL canonical, permissive read / preserving write) + additive remote_id

usercise:
## Summary

Freezes the git-notes v1 format (`refs/notes/spelunk`) and adds an optional, additive cross-machine id.

We are not the only writer of git notes, so the format is permissive on read and strict on preservation when writing:

- **Permissive read.** A commit's note is JSON Lines, one `NoteRecord` per line, possibly interleaved with foreign content (prose, other tools' lines). Reads parse the lines we understand and skip blank, non-JSON, non-object, and foreign lines without erroring. Only a record from a newer, incompatible `schema_version` returns an error. This removes the old failure where one multi-entry or unparseable line poisoned the whole `memory list --backend git`.
- **Preserving write.** `add` and `archive` route through a read-modify-write helper that re-emits every existing line (sibling records and foreign prose) byte-for-byte and mutates only the target record. The previous single-record force-replace that silently deleted siblings and foreign content is gone.
- **Additive `remote_id`.** A new optional `remote_id` (uuid string) on `NoteRecord`, the server `notes` table (new forward migration `server_004.sql`, nullable column plus a partial UNIQUE index), and the `/v1` `NoteResponse` / `AddNoteResponse` wire types. It is the canonical cross-machine id, set on sync to a remote server. Fully backward-compatible: absent means `None`, old readers ignore it, and the local i64 `id` is untouched.

Implements ADR-059.

## Test plan

- Merged `origin/main` into the branch to de-stale it (the branch had forked before the `add_note` audit-log injection-rejection change landed). The merge was clean with no conflicts; that change and this branch's changes touch disjoint files, so both behaviors are retained.
- `SPELUNK_SECRET_STORE=file cargo test --workspace` — all suites green, 0 failures. Includes `spelunk-core` lib (216), git-notes integration (16), server unit (71), and server integration (12).
- `cargo fmt --check` — clean.
- `cargo clippy --all-targets` — clean, no warnings.
- Conformance fixture from ADR-059 (the "Implement payment by Stripe" note with markdown headers, freeform prose, and three `{ "kind": "decision", ... }` lines) is exercised by the git-notes integration tests:
  - reading yields exactly the three records in order, prose skipped, no error;
  - a foreign-only blob (prose plus a non-object JSON value plus a malformed line) reads as an empty list with no error;
  - `add` retains all prose and siblings and appends a fourth record;
  - `archive` of the middle record mutates only that record, leaving siblings and prose unchanged in content and position.
- A dedicated server test confirms re-opening the DB re-runs the non-idempotent `ALTER TABLE ADD COLUMN` without failing and that existing rows default to `remote_id` NULL.

## Backward-compatibility guarantees

- Absent `remote_id` reads as `None`; old blobs and old wire payloads without the key deserialize unchanged.
- Old readers ignore the new key.
- The local i64 `id` is unchanged.
- The permissive read never fails on foreign or unparseable lines; only `schema_version > 1` errors.
- Preserving writes never clobber sibling records or foreign content.

Closes spelunk-oss^104 and spelunk-oss^105.


crates/spelunk-core/src/storage/remote/wire_types.rs
```diff
@@ -51,6 +56,11 @@ pub(super) struct NoteResponse {
     pub(super) valid_at: Option<i64>,
     #[serde(default)]
     pub(super) invalid_at: Option<i64>,
+    /// Canonical cross-machine id, if the server has one. Absent on older
+    /// servers → `None`. Accepted for forward-compat; not surfaced into `Note`.
+    #[serde(default)]
+    #[allow(dead_code)]
```
usercise:
```suggestion
```

Why do we need allow dead code in a new data struct?

crates/spelunk-core/src/storage/remote/wire_types.rs
```diff
@@ -23,6 +23,11 @@ pub(super) struct AddNoteResponse {
     pub(super) id: i64,
     #[serde(default)]
     pub(super) conflicts: Vec<ConflictInfo>,
+    /// Server-assigned cross-machine id, if the server minted one. Absent on
+    /// older servers → `None`.
+    #[serde(default)]
+    #[allow(dead_code)]
```
usercise:
```suggestion
```

crates/spelunk-core/tests/integration_git_notes.rs
```diff
@@ -493,3 +494,199 @@ async fn git_notes_backend_add_with_option_like_body_round_trips() {
...
+fn write_raw_note(root: &std::path::Path, blob: &str) {
+    use std::io::Write;
+    use std::process::{Command, Stdio};
+    let head = String::from_utf8(
+        Command::new("git")
+            .args(["-C", root.to_str().unwrap(), "rev-parse", "HEAD"])
+            .output()
+            .expect("rev-parse")
+            .stdout,
+    )
```
usercise, line 515: Are we sure we can't get HEAD sha from gix? Possibly here https://github.com/GitoxideLabs/gitoxide/blob/main/gix-refspec/src/parse.rs

crates/spelunk-core/src/storage/remote/wire_types.rs
usercise: Removed. The field was dead only because nothing consumed it. Now `RemoteMemoryBackend::add` reads `AddNoteResponse.remote_id` and logs the server-minted id (`tracing::debug!`) — this backend has no local store to persist it into, so surfacing it for diagnostics is the honest minimal consumption per ADR-059 D2 ("so a client can record the server-assigned remote_id on write"). `#[allow(dead_code)]` gone; clean under both feature configs.

crates/spelunk-core/src/storage/remote/wire_types.rs
usercise: Removed. `remote_id` now flows into the domain `Note`: added `remote_id: Option<String>` to `Note` and populate it in `From<NoteResponse> for Note` (and in `record_to_note` from the git-notes `NoteRecord`). It is read via `Note`'s `Serialize` output, so the field is genuinely consumed and `#[allow(dead_code)]` is gone. This is the additive plumb ADR-059 D2 intended; no notes-as-sync consumer added (that stays a post-v1 non-goal). Green under both feature configs.

crates/spelunk-core/tests/integration_git_notes.rs
usercise: Left as-is (shell-out), and here's why. The production git-notes backend is entirely `git`-subprocess: `append_to_git_notes` gets HEAD via `git rev-parse HEAD` (`git_notes/mod.rs:33`), `GitNotesBackend::head_sha` does the same (`backend_impl` path), and every note read/write is a `git notes` subprocess (there is no gix note-write API in use). This test helper deliberately mirrors that exact path — `git rev-parse HEAD` then `git notes --ref=spelunk add -F -`. Switching only the HEAD resolution to `gix::open(root)?.head_id()` would introduce a single, inconsistent gix call into an otherwise 100%-subprocess subsystem, and the test would no longer exercise the same resolution the production write-through uses. (Side note: the linked `gix-refspec/parse.rs` is refspec parsing, not HEAD resolution — the relevant API would be `Repository::head_id`.) If/when the backend itself moves off subprocess to gix, the helper should move with it in the same change so test and production stay consistent.

PR #538, opened 2026-07-07
docs: rewrite OSS user docs to the approved journey, grounded in code (spelunk-oss^12)

usercise:
## What / why

Rewrites the OSS user docs to the founder-approved journey spine (decision #275): on-ramp (fast codebase understanding) -> getting started -> configure your agent -> reference -> local vs server vs team-server. The on-ramp intentionally hooks on "how" (fast understanding); "why" is the earned payoff, and the "ask me why" brand line is preserved. Every command, flag, and config key touched is verified against the shipped binary rather than from memory.

## Key changes

- New journey index at `docs/README.md` that walks the path from the first five minutes to running a shared memory server; root `README.md` now points at it.
- `spelunk ask` (never implemented in OSS) purged everywhere and, where a real command was meant, replaced with `spelunk explore`.
- `sync` corrected from "alias for `memory push`" to genuine two-way sync (push + pull), matching the `Sync` / `MemorySyncArgs` definitions; `memory pull` and `memory sync` documented alongside `memory push`.
- Config `batch_size` key removed (it is not a config key); documented as the real `spelunk index --batch-size <n>` flag instead.
- Tier model corrected across `getting-started.md` and `docs/README.md`: auto-discovered loopback server = inference only / never a memory store; local `memory.db` = default store; explicit `server_url` = the only path memory leaves the machine.
- `docs/example/` consolidated into `docs/examples/` (AGENT.md moved; `spelunk ask` and em-dashes cleaned in the template).
- `docs/mcp.md` removed (MCP is a "(Future)" placeholder, unimplemented in OSS) and `docs/plans/*` removed (stale internal roadmaps of unbuilt features).

## Test plan

Grounding spot-checks against source (all matched):
- Top-level verbs (`init`, `index`, `search`, `context`, `graph`, `memory`, `hooks`, `explore`, `sync`, `server`, `login`) present in the clap `Command` enum (`crates/spelunk-cli/src/cli/mod.rs`).
- `Sync` doc-comment = "Two-way sync ... (shorthand for `memory sync`)"; `SyncArgs` is a re-export of `MemorySyncArgs`.
- `memory push` (one-way) / `pull` (server->local) / `sync` (two-way) all present in `MemoryCommand` with matching semantics.
- `--batch-size` is a real flag on `index` (`cli/cmd/index/mod.rs`); no `batch_size` config key in `crates/spelunk-core/src/config.rs`.
- `spelunk sync` flags `--project` / `--source` / `--include-archived` match `MemorySyncArgs`.
- `memory harvest --git-range` and `--source claude-code` are real flags (`MemoryHarvestArgs`); the `claude-code` example now includes the required `--confirm`.
- `hooks install` exists and the installed post-commit hook runs `spelunk index` + `spelunk memory harvest`.
- plumbing subcommands (`cat-chunks`, `parse-file`, `embed`, `knn`, `hash-file`, `graph-edges`) all exist as real handlers.
- `grep -rniE "spelunk ask" docs README.md` -> zero.
- Link/anchor check: every link in `docs/README.md` resolves (target file + `#anchor`); `example/`->`examples/` left no dangling links; `capability-tiers.md` updated to the renamed anchor.
- `git diff main...HEAD | grep '^+' | grep '-'` (em-dash) -> empty; public-boundary term sweep on added lines -> clean.

## Follow-ups

- `docs/architecture/capability-tiers.md` deserves its own grounding pass (spelunk-oss follow-up task).
- Deleted `docs/plans/*` are dropped but recoverable from git history if any turn out to be wanted.
- `CLAUDE.md` (repo root, outside `docs/`) still calls `sync` an "alias for memory push" in two places - worth a separate cleanup.
- `docs/server.md` still links the old `#server-mode-vs-no-server-mode` anchor; that file belongs to spelunk-oss^101 and was left untouched (already flagged there).
- `docs/security/THREAT-MODEL.md` still cites `ask.rs` (a source file that no longer exists; logic lives in `explore.rs`) on pre-existing lines this rewrite did not touch - candidate for a later grounding pass.


docs/README.md
```diff
@@ -0,0 +1,89 @@
...
+- [Memory](memory.md): decisions, requirements, and context; supersede, do not delete
+
+## 3. Configure your agent
+
+Wire an AI coding agent to spelunk so it pulls context before it edits and
+records decisions as it works. The payoff is that the why-layer fills itself:
+install the git hook and every commit runs `spelunk memory harvest`, so the
+reasoning behind the code is captured automatically, with no separate step to
+sit down and write docs. This is what completes the how-to-why arc: the decisions
+stage 1 could not yet show now accumulate on their own.
```
usercise: This feels like agent instructions, not user information.

```suggestion
sit down and write docs.
```

docs/README.md
```diff
@@ -0,0 +1,89 @@
...
+
+- [Commands](commands.md): every subcommand, flag, and environment variable
+- [Memory model](memory.md): kinds, cross-project visibility, git-notes write-through
+- [Architecture](architecture.md) and [capability tiers](architecture/capability-tiers.md)
+- [Plumbing and porcelain](plumbing-and-porcelain.md): JSONL commands for scripts and agents
+- [Security](security/THREAT-MODEL.md): threat model and boundaries (secret scanning is defense-in-depth, not a boundary)
+
+## 5. Local vs server vs team-server
+
+The three tiers, stated once. A local server does inference only; an explicit
```
usercise: Again agent instruction not user information.

```suggestion
Too same speed and memory we run embedding in a local server, so the context don't disappear with each cli invocation. That is all the local server does, it doesn't store memory and it is not accessible from outside your machine. For sharing memory we provide the team server (also open source and free), or our cloud package if you prefer a managed solution. The team server shares your memory (but not code) index so that the whole team can work of the same source of truth.
```

docs/README.md
```diff
@@ -23,27 +23,33 @@ where, what), not a faster grep, and it needs no server.
...
+This is where the payoff lands. If you code with an AI agent, you connect it to
+spelunk once and the why-layer starts filling itself: as the agent works, the
+reasoning behind each change is captured for you, with no time set aside to sit
+down and write it up. The decisions a fresh repo could not yet show in stage 1
+now accumulate on their own, and every later `spelunk context` or `spelunk
+search` hands them back.
 
-- [Agent Guide](agent-guide.md): the session pattern, automatic capture, and machine-readable output
-- [AGENT.md template](examples/AGENT.md): drop-in instructions telling an agent to use spelunk
+The mechanism is a single git hook. Installing it adds a post-commit step that
```
usercise: "The mechanism is a single git hook" - is it though? The git hook helps, but what we are getting at this level is the back pressure of the agent recording the decisions as it goes along, via a skill.

docs/README.md
```diff
@@ -61,29 +67,39 @@ before it.
...
+| Team memory server (explicit `server_url`) | one shared memory index for the team | the shared server: the only path off the local machine |
 
-Point a team at a self-hosted spelunk server with an explicit `server_url` and a
-shared server key, then use [`spelunk sync`](commands.md#spelunk-sync) to push
-your decisions up and pull teammates' decisions down. Each developer's code stays
-local.
+Everyone on a team sets an explicit `server_url` (plus a shared server key)
+pointing at the same server, and [`spelunk sync`](commands.md#spelunk-sync) keeps
+them converged: it pushes the decisions you recorded and pulls the ones your
+teammates recorded. Code never travels; only memory does.
```
usercise: Code never travels, it doesn't need to, you have git for that; only memory does

PR #546, opened 2026-07-08
docs: fix advanced-inference section (api_base_url is not a config key)

usercise:
## What

The **Using your own inference server (advanced)** section in `docs/getting-started.md` documented a CLI `config.toml` mechanism that does not exist.

### The drift

- `api_base_url` is **not** a field on the `Config` struct (`crates/spelunk-core/src/config.rs`) and has no serde alias, so setting it in `config.toml` is silently ignored (`Config` does not `deny_unknown_fields`). The nearest real field, `inference_url`, is `#[serde(skip)]` — runtime-only, populated by tier discovery, never read from config files.
- `embedding_model` exists but is only a **display label** for `plumbing embed` JSONL output ("The effective embedding model is owned by spelunk-server, not this key"). Its real default is `f2llm-v2-330m`, not the `text-embedding-embeddinggemma-300m-qat` shown.
- All inference routes through **spelunk-server**, so a custom OpenAI-compatible endpoint is a *server* setting, not a CLI one.

### The fix

Rewrote the section to use the actual mechanism, **verified against `spelunk-server --help`** (`crates/spelunk-server/src/main.rs`):

| Variable | Flag |
|---|---|
| `SPELUNK_EMBEDDING_URL` | `--embedding-url` |
| `SPELUNK_EMBEDDING_MODEL` | `--embedding-model` |
| `SPELUNK_LLM_URL` | `--llm-url` |
| `SPELUNK_LLM_MODEL` | `--llm-model` |

Also:
- Documented the embedding-dimension constraint (a non-native embedder differs from the native 896-dim; set `--embedding-dim` and start from a fresh index — the server records the dim on first write and rejects mismatches), cross-linked to `server.md#embedding-dimension`.
- Kept the real `spelunk index --batch-size <n>` flag (verified it exists on `IndexArgs`), just tightened wording to "ceiling".
- Fixed the follow-on cross-reference that said "pointed spelunk at your own inference server".

## Why

Surfaced while mirroring OSS docs into the marketing site; the marketing config-reference already omits `api_base_url` for this reason. Once this lands, the marketing getting-started can add back a correct advanced-inference note.

## Notes

- Docs-only change; `cargo fmt --check` + `cargo clippy` ran clean via pre-commit hook.
- Out of scope but related: `docs/adr/002-server-ai-endpoints.md` still references `api_base_url` as if it works for `explore`/`index` — that's a historical ADR, left untouched.

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -264,23 +264,52 @@ For how discovery works and how to point the CLI at a remote server, see
 
 By default the bundled `spelunk-server` provides embeddings (native, via the
 candle-served F2LLM-v2-330M model) and — when a chat model is configured — LLM
-inference. If you'd
-rather have spelunk talk directly to your own OpenAI-compatible endpoint (e.g.
-LM Studio on port `1234`, Ollama on `11434`, or a vLLM proxy) instead of the
-native embedder, point it there in `~/.config/spelunk/config.toml`:
+inference. If you'd rather have it call your own OpenAI-compatible endpoint
+(e.g. LM Studio on port `1234`, Ollama on `11434`, or a vLLM proxy) instead of
+the native embedder, configure **the server** — this is not a CLI
+`config.toml` key. `spelunk-server` reads these environment variables (each has
+an equivalent flag):
 
-```toml
-# ~/.config/spelunk/config.toml
+| Variable | Flag | Purpose |
+|---|---|---|
+| `SPELUNK_EMBEDDING_URL` | `--embedding-url` | Base URL of an OpenAI-compatible embedding endpoint. When set, the server embeds through it instead of the native model. |
+| `SPELUNK_EMBEDDING_MODEL` | `--embedding-model` | Model id to send to that endpoint (must match what it serves). |
+| `SPELUNK_LLM_URL` | `--llm-url` | Base URL of an OpenAI-compatible chat-completions endpoint for LLM features (`explore`, summaries, `memory harvest`). |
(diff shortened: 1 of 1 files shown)


docs/getting-started.md
```diff
@@ -264,32 +264,61 @@ For how discovery works and how to point the CLI at a remote server, see
...
+inference. If you'd rather have it call your own OpenAI-compatible endpoint
+(e.g. LM Studio on port `1234`, Ollama on `11434`, or a vLLM proxy) instead of
+the native embedder, configure **the server** — this is not a CLI
+`config.toml` key. `spelunk-server` reads these environment variables (each has
+an equivalent flag):
+
+| Variable | Flag | Purpose |
+|---|---|---|
+| `SPELUNK_EMBEDDING_URL` | `--embedding-url` | Base URL of an OpenAI-compatible embedding endpoint. When set, the server embeds through it instead of the native model. |
+| `SPELUNK_EMBEDDING_MODEL` | `--embedding-model` | Model id to send to that endpoint (must match what it serves). |
```
usercise, line 276: Embedding models should no longer be possible to override.

usercise: Tracked as a separate task: **spelunk-oss^119** — remove the embedding-model override (`--embedding-model` / `SPELUNK_EMBEDDING_MODEL`) from `spelunk-server`, since the embedding model is fixed product-wide post-ADR-053 (ADR-010 one-model-per-vector-space invariant). Left out of this PR so #546 stays a pure `api_base_url` docs fix; the override removal + its own docs update live in oss^119.

PR #559, opened 2026-07-09
spelunk-oss^120: fix macOS MDM example to bind loopback (--host 127.0.0.1)

usercise:
Fixes the shipped macOS MDM example (spelunk-oss^120). The LaunchDaemon
`ProgramArguments` bound `--host 0.0.0.0`, which `check_bind_safety` rejects,
so the daemon would fail to start as shipped.

## Changes (data-only, one file)
- `examples/mdm/macos/cloud.spelunk.server.mobileconfig`: `--host 0.0.0.0` →
  `--host 127.0.0.1`; adds a header "REACHING IT OFF-HOST" note and an inline
  `ProgramArguments` comment pointing operators at an operator-owned TLS reverse
  proxy (Caddy, nginx) forwarding HTTPS to `127.0.0.1:7777`. This matches the
  Windows MDM example and `docs/self-hosting.md`. No Rust changed.

## Test plan
- `plutil -lint examples/mdm/macos/cloud.spelunk.server.mobileconfig` → OK.
- No em-dashes in the added copy (public repo).
- Built `spelunk-server`; launched it with the example's corrected LaunchDaemon
  args (`--host 127.0.0.1 --port 7790 --db <tmp>`, `SPELUNK_SERVER_KEY` set):
  server starts, logs `spelunk-server listening on http://127.0.0.1:7790`, and a
  `--health-check` probe against it exits 0. No `check_bind_safety` refusal.
- Sanity check: the OLD args (`--host 0.0.0.0` + key) still fail-fast:
  `Refusing to bind to non-loopback address '0.0.0.0:7789' over plaintext HTTP:
  the bearer SPELUNK_SERVER_KEY would cross the network in cleartext.` (exit 1).

### DIFF
--- a/examples/mdm/macos/cloud.spelunk.server.mobileconfig
+++ b/examples/mdm/macos/cloud.spelunk.server.mobileconfig
@@ -20,12 +20,11 @@
   see ../README.md ("Pushing spelunk configuration").
 
   REACHING IT OFF-HOST
-  This job binds loopback (--host 127.0.0.1). spelunk-server refuses a
-  non-loopback plaintext bind with no override, so do NOT change --host to
-  0.0.0.0 (the daemon would fail to start). To serve other machines, run an
-  operator-owned TLS reverse proxy (Caddy, nginx) on this host forwarding
-  HTTPS to 127.0.0.1:7777, exactly like the Windows example and the
-  bare-metal shape in ../../../docs/self-hosting.md.
+  This job binds loopback (--host 127.0.0.1) only, and this profile is scoped
+  to local, on-host use. spelunk-server refuses a non-loopback plaintext bind
+  with no override, so do NOT change --host to 0.0.0.0 (the daemon would fail
+  to start). This profile does not expose the server to other machines;
+  reaching it off-host is a separate deployment decision, out of scope here.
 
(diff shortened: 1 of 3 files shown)


examples/mdm/macos/cloud.spelunk.server.mobileconfig
```diff
@@ -19,6 +19,14 @@
   deploy the config file and/or environment via an MDM file or script payload
   see ../README.md ("Pushing spelunk configuration").
 
+  REACHING IT OFF-HOST
+  This job binds loopback (--host 127.0.0.1). spelunk-server refuses a
+  non-loopback plaintext bind with no override, so do NOT change --host to
+  0.0.0.0 (the daemon would fail to start). To serve other machines, run an
+  operator-owned TLS reverse proxy (Caddy, nginx) on this host forwarding
```
usercise: I've said this multiple times already - we will not recommend that people put a proxy in front of the server. 

There is a explicit security boundary:

- Local only is http only and does not require authentication (no api key)
- Remove is https only and requires an api key to be set.

By recommending a proxy in front of the local server we are breaking that security boundary.

PR #562, opened 2026-07-10
ADR-066: native in-process HTTPS for spelunk-server (reverses ADR-058 Non-goal)

usercise:
**Draft ADR for founder approval. Do not merge before Johan signs off** — this reverses a founder-approved Non-goal in ADR-058.

Refines spelunk-oss^123. Per Johan's 2026-07-10 directive: *local = HTTP loopback, no key; remote = HTTPS, API key required; we do not recommend putting a proxy in front of the server.* Today `spelunk-server` has no TLS and `check_bind_safety` refuses every non-loopback bind, so the only off-host path is an operator proxy in front of a loopback bind — exactly what that directive rejects. This ADR gives the server its own HTTPS.

## What it decides
- **TLS via rustls + `axum-server`** (`from_tcp_rustls`, preserving the bind-before-embedder-warm ordering). rustls is already in the tree via reqwest/hf-hub.
- **Crypto provider pinned to `ring`** (the provider already resolved in the tree) to avoid pulling `aws-lc-sys` — which would add cmake + a C compiler + NASM-on-Windows as a new build dependency. Validated by a compile spike (`tls-rustls-no-provider` + `ring`): zero `aws-lc-sys`.
- **BYO PEM cert** via `--tls-cert` / `--tls-key` (deliberately not `--key`, which is the bearer API key). **ACME deferred** to a possible future ADR.
- **`check_bind_safety` becomes TLS-aware**: a non-loopback bind is allowed only when TLS **and** a key are both set; plaintext off-host stays refused with no opt-out. Nothing currently permitted is removed (keyed-plaintext off-host was already refused by spelunk-oss^122).
- **Private TLS key** gets the same systemd `LoadCredential=` treatment as the bearer key.
- **Revisits ADR-058's Docker exclusion**: that exclusion rested solely on loopback-only + external proxy. A routable in-process TLS bind makes a container reachable, so Docker returns as a supported team-server shape. Bare-metal + systemd stays the reference recipe (now binding the routable interface with TLS instead of loopback + proxy).
- **ADR-056 (tenancy) is orthogonal** and unchanged: the shared key is still the only tenancy boundary; TLS just keeps it off the wire.

## Also in this PR
- Supersede note at the top of ADR-058 + annotation of the "Native TLS" Non-goal bullet (history preserved, not rewritten).

## Spike evidence
Throwaway compile spike (axum 0.8.9 + axum-server 0.7.3 + rustls 0.23 `ring`): `RustlsConfig::from_pem_file`, `from_tcp_rustls(pre_bound_listener, cfg)`, and `into_make_service_with_connect_info::<SocketAddr>()` all type-check; ring path pulls no `aws-lc-sys`. Spike is not shipped.

## Not in scope
Implementation, tests, and the docs/MDM restructure (`self-hosting.md`, `server.md`, `examples/mdm/*`, `packaging/*.service`, `remote-agents.md`) are planned in §7 of the ADR and land in a separate PR **after** approval.

### DIFF
--- a/docs/adr/066-native-tls-in-spelunk-server.md
+++ b/docs/adr/066-native-tls-in-spelunk-server.md
@@ -255,5 +255,3 @@ bearer key in ADR-058:
 - rustls with the `ring` provider and safe protocol defaults (TLS 1.2+/1.3)
   keeps the added stack memory-safe and conservatively configured; no
   operator-tunable cipher policy is exposed in this iteration.
-</content>
-</invoke>
--- a/docs/adr/066-native-tls-in-spelunk-server.md
+++ b/docs/adr/066-native-tls-in-spelunk-server.md
@@ -1,0 +1,259 @@
+# ADR-066: Native in-process HTTPS for spelunk-server
+
+**Date:** 2026-07-10
+**Deciders:** founder (Johan), architect
+**Supersedes:** the "Native TLS in spelunk-server" Non-goal in
+[ADR-058](058-team-server-bare-metal-deployment.md), and revisits that ADR's
+Docker-vs-bare-metal reasoning (both were predicated on the server having no
+in-process TLS, which this ADR removes).
+
+## Context
+
+`spelunk-server` speaks plaintext HTTP only. It has no TLS: no `--cert`, no
+`--tls-*` flags, no in-process terminator. Its one transport guard,
+`check_bind_safety` (`crates/spelunk-server/src/main.rs`), refuses **every**
+non-loopback bind, whether keyless (an open server) or keyed (the bearer key
+would cross the wire in cleartext). That refusal is unconditional and has no
+opt-out (hardened in spelunk-oss^122). Loopback binds are always allowed.
+
+The consequence: the only code-supported way to reach the server from another
+machine is to bind loopback and put an operator-owned TLS reverse proxy in front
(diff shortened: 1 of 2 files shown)


docs/adr/066-native-tls-in-spelunk-server.md
```diff
@@ -0,0 +1,259 @@
...
+- The TLS **private key** is treated as a high-value secret, mirroring the bearer
+  key: supplied via a systemd credential (or a `0600` root-owned file), never an
+  `Environment=` line, kept out of `systemctl show` / `/proc/<pid>/environ`.
+- The tenancy boundary is unchanged (ADR-056): the shared key still grants full
+  admin of every project on the instance. TLS protects the key in transit; it
+  does not add per-project isolation.
+- rustls with the `ring` provider and safe protocol defaults (TLS 1.2+/1.3)
+  keeps the added stack memory-safe and conservatively configured; no
+  operator-tunable cipher policy is exposed in this iteration.
+</content>
```
usercise: The last two lines are agent structure, don't belong in the ADR.

usercise: Fixed in 30e40ce42 — stripped the stray `</content></invoke>` tool-call tags; the ADR now ends at the cipher-policy line. Nothing else touched.

PR #601, opened 2026-07-13
feat(cli): trust self-signed / internal-CA team-server certs via SPELUNK_SERVER_CA

usercise:
## What & why

Adds CLI trust for team servers whose TLS certificate is signed by a self-signed
or internal (corporate) CA, so agents behind a corp firewall can reach a
self-hosted `server_url` without a publicly-trusted cert. This is additive to
ADR-066 (native-TLS team server): the transport, API-key auth, and non-loopback
guard are unchanged; this only teaches the client which extra CA to trust.

## Trust surface

- New env var `SPELUNK_SERVER_CA` and config key `server_ca` (in
  `.spelunk/config.toml`) point at a PEM CA bundle. **Env overrides config.**
- The bundle is added as a trust anchor **on top of** the built-in roots via
  `config::apply_server_ca`. Certificate verification stays **ON** — there is no
  `--insecure` / `danger_accept_invalid_certs` / disable switch anywhere in
  shipped code.
- The path routes through every team-server client site (inference, capability
  probe, remote memory backend + sync, `memory since` / `memory watch`, index
  embed phase). Cloud-API and loopback-daemon clients are unaffected.

## Caveat (CA vs bare leaf)

The trusted material must be a proper **CA** certificate (or a CA-to-leaf
chain). A bare self-signed `openssl req -x509` certificate is `CA:TRUE` and
rustls rejects it if the server presents it as its own end-entity cert
(`CaUsedAsEndEntity`). Operators must mint an internal CA and issue the server a
`CA:FALSE` / `serverAuth` leaf. This is documented in `docs/self-hosting.md`,
`docs/commands.md`, and the CHANGELOG.

## native-roots deferred

OS/native trust-store auto-discovery (`rustls-tls-native-roots`) was
intentionally **not** included: it deterministically broke
`secret_scanner::summary_secret_is_not_persisted` via a summary-thread race.
That is deferred pending a separate fix; the explicit CA-bundle path here has no
such interaction.

## Test plan (actually run in this re-gate)

- `secret_scanner::summary_secret_is_not_persisted` standalone x3 → 3/3 GREEN
  (was 3/3 red before native-roots was dropped).
- `SPELUNK_SECRET_STORE=file cargo test -p spelunk-cli` → green.
- `SPELUNK_SECRET_STORE=file cargo test` (full workspace, default features) →
  green, including `spelunk-server --test tls_serve` (3 passed): the real
  self-signed round-trip plus the internal-CA leaf round-trip with an
  untrusted-client negative control proving verification stays on.
- `cargo build --no-default-features` → clean (exit 0).
- `cargo clippy --all-targets` → clean (exit 0).
- `cargo fmt --check` → clean.

Note: CHANGELOG may need a trivial rebase against sibling PRs at merge time.

### DIFF
--- a/crates/spelunk-cli/src/cli/cmd/memory/push.rs
+++ b/crates/spelunk-cli/src/cli/cmd/memory/push.rs
@@ -51,7 +51,12 @@ pub async fn memory_push(
         .with_context(|| format!("opening local memory at {}", src_path.display()))?;
     // Refresh a stale WorkOS access token before the cloud-api call (oss^40).
     let key = auth_api::ensure_fresh_server_key(cfg).await?;
-    let client = CloudSyncClient::new(&base_url, &project_id, key.as_deref())?;
+    let client = CloudSyncClient::new(
+        &base_url,
+        &project_id,
+        key.as_deref(),
+        cfg.server_ca.as_deref().map(std::path::Path::new),
+    )?;
 
     println!("Pushing local memory to {base_url}…");
     let summary = push_local_oneway(&local, &client, args.include_archived).await?;
--- a/crates/spelunk-core/src/storage/remote/mod.rs
+++ b/crates/spelunk-core/src/storage/remote/mod.rs
@@ -139,6 +139,7 @@ pub async fn resolve_cloud_project_uuid(
     project_id: &str,
     server_url: &str,
     server_key: Option<&str>,
+    server_ca: Option<&std::path::Path>,
     spelunk_dir: &std::path::Path,
 ) -> Result<uuid::Uuid> {
     // D5: raw UUID in config → use directly, no lookup, no cache.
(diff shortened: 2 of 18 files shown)


crates/spelunk-cli/src/cli/cmd/memory/push.rs
```diff
@@ -51,7 +51,12 @@ pub async fn memory_push(
         .with_context(|| format!("opening local memory at {}", src_path.display()))?;
     // Refresh a stale WorkOS access token before the cloud-api call (oss^40).
     let key = auth_api::ensure_fresh_server_key(cfg).await?;
-    let client = CloudSyncClient::new(&base_url, &project_id, key.as_deref())?;
+    let client = CloudSyncClient::new(
```
usercise, line 54: Cloud does not need internal-CA support, only team server needs it.

crates/spelunk-core/src/storage/remote/mod.rs
```diff
@@ -158,7 +159,8 @@ pub async fn resolve_cloud_project_uuid(
         );
     }
 
-    resolve_cloud_project_uuid_inner(project_id, server_url, server_key, spelunk_dir).await
+    resolve_cloud_project_uuid_inner(project_id, server_url, server_key, server_ca, spelunk_dir)
```
usercise, line 162: Same, if this is cloud only we don't need ca support here.

PR #582, opened 2026-07-12
feat(init): configure origin notes refspec so memory travels (spelunk-oss^126, ADR-068)

usercise:
## What

`spelunk init` now configures the `origin` **fetch** refspec so project memory
(`refs/notes/spelunk`) travels on `git fetch`. When init runs inside a git repo
with an `origin` remote, it idempotently adds
`+refs/notes/spelunk:refs/notes/spelunk` to `remote.origin.fetch` (checks
`--get-all` first, no duplicate on re-run) and announces the change in the init
summary. It never writes git config silently. (spelunk-oss^126, ADR-068)

## Fetch-only refspec, no push refspec — and why

The change deliberately does **not** set `remote.origin.push`. Any
`remote.origin.push` value overrides git's default branch push, so a normal
`git push` would stop pushing the current branch. To keep the branch-push
default intact, init instead surfaces the one-time manual publish command:

```
git push origin refs/notes/spelunk
```

- No `origin` remote or non-TTY: graceful skip, exit 0, no prompt/hang. In the
  no-origin case init prints the exact `git config --add …` + push commands to
  run later.
- Uses `refs/notes/spelunk` (spelunk's ref), not `refs/notes/commits`.

## Docs

`docs/memory.md`, `docs/commands.md`, and `CHANGELOG.md` document the flow and are
careful **not** to overstate: notes do not travel automatically on `git clone`
(a fresh clone does not inherit the source's local git config). Teammates must
run `spelunk init` in their clone (or add the refspec manually) and then
`git fetch`; the publisher pushes once. The marketing-site getting-started
wording is a separate follow-up, not in this PR.

## Test plan (actual steps run in this review)

- `cargo test -p spelunk-cli` → all green, including the 6 new tests in
  `tests/init_notes_refspec.rs` (origin-present config, no-origin hint,
  idempotency, `remote.origin.push` stays unset, non-TTY no-hang, and a
  bare-origin round-trip).
- `cargo test -p spelunk-cli --no-default-features` → all green (new tests pass).
- `cargo clippy --all-targets` → clean, no warnings.
- Round-trip test (`notes_round_trip_through_bare_origin`) verified to be a real
  assertion, not a no-op: it pushes a `memory add` decision note to a bare
  origin, clones fresh, asserts the clone has **no** spelunk notes pre-fetch,
  runs init + `git fetch` in the clone, then `git notes --ref=spelunk show`s the
  annotated object and asserts the decision **title/content** arrived.

[...]

PR #604, opened 2026-07-14
docs(adr): ADR-069 share spelunk memory on git push via an opt-in hook + tracking refspec

### LINKED ISSUE #185: Spike: concurrent write safety test for git notes backend
## Context

Part of #181 (git notes memory backend spike).

The research note states "merge conflicts on notes are rare." That is true for human writes, but spelunk's primary consumer is AI agents writing concurrently. If two agents call `git notes --ref=spelunk add HEAD` simultaneously, the second overwrites the first — `git notes add` uses replace semantics by default, not append. The `-f` flag makes this unconditional. A multi-agent workflow (e.g. two Claude Code sessions open on the same repo) can silently lose memory entries.

This needs to be tested and the outcome documented before `GitNotesBackend` is considered safe for agent use.

## Task

Write a test script (or Rust integration test) that:

1. Initialises a temporary git repo with one commit
2. Spawns two concurrent processes, each running `spelunk memory add --backend git-notes --kind note --title "agent A/B" --body "..."`
3. After both complete, runs `spelunk memory list --backend git-notes` and asserts **both** entries are present

Then investigate mitigations:

**Option A — One note per commit, array of entries**
Store a JSON array in the note rather than a single object. `add` reads the current note, appends, and rewrites atomically using a file lock on `.git/spelunk.lock`. Pros: single note per commit. Cons: file locking is not network-safe (NFS).

**Option B — One note per entry, keyed by timestamp**
Use `git notes --ref=spelunk/<timestamp-nanos>` so each entry has its own ref. `list` walks all `refs/notes/spelunk/*`. Pros: no conflict. Cons: ref namespace bloat; `git push` requires explicit glob.

**Option C — Accept last-write-wins, document it**
Document that concurrent agent writes may lose entries and advise single-agent use for git-notes backend. Acceptable tradeoff for a v1 spike.

## Deliverable

A written assessment (comment on this issue or PR description) covering:
- Which option was implemented in the spike
- Whether the team round-trip test still passes after the fix
- Whether the chosen approach is acceptable for the OSS v1 or needs more work before promotion to default

## Success criteria

- Concurrent write test is documented (pass or known-fail with explanation)
- The chosen conflict strategy is recorded as a `spelunk memory add --kind decision` entry

usercise:
Records the decision from the git-notes sharing spike. Resolves the open question [ADR-068](https://github.com/spelunk-cloud/spelunk/blob/main/docs/adr/068-zero-setup-onboarding-git-notes-memory-fallback.md) deferred to ^126: *"Does 'travels with the repo' require configuring the notes refspec?"*

The answer is yes, and the mechanism currently on `main` (#582) is wrong in two ways that this ADR corrects.

## What it decides

- **D1 – sharing is coupled to `git push`,** via an opt-in pre-push hook. Not to `memory add`, not to a timer. A note on an unpushed commit reaches origin while its target object does not, so the memory is orphaned and teammates never see it. `git push` is the only moment that coincides with "this code is being shared." It also sidesteps the `remote.origin.push` constraint #582 correctly identified: a hook pushes the notes ref separately and leaves the branch-push default intact.
- **D2 – `cat_sort_uniq` is the canonical merge strategy.** spelunk appends each entry as a JSON line, so a note is a union set. `ours`/`theirs` exit 0 while silently destroying one side's memory; the default (`manual`) exits 1 and leaves a stuck partial merge. Only `cat_sort_uniq` unions cleanly, and it is never the default, so it must be passed explicitly.
- **D3 – opt-in, non-blocking, best-effort.** The hook must never block the user's push (a hook exiting 1 aborts the branch push). The recursion guard is mandatory: a naive version recursed 740 levels and stopped only by exhausting the process table, with outer pushes failing invisibly.
- **D4 – correct the fetch refspec to a tracking ref.** `+refs/notes/spelunk*:refs/notes/origin/spelunk*`.

## Why D4 matters most

#582's `+refs/notes/spelunk:refs/notes/spelunk` has two independent defects on git 2.55.0:

1. The non-glob form requires the remote ref to exist, so with no notes on the remote (every repo until someone pushes) `git fetch origin` exits 128 and `git pull` exits 1. **`spelunk init` breaks the user's normal git workflow today.**
2. The leading `+` on a working ref force-updates it: a local unpushed note was silently replaced by the remote's on a plain `git fetch`, recoverable only via reflog. That is data loss of the product's core asset.

A glob alone fixes (1) but not (2); only the tracking destination is safe.

## Notes for review

- Two consequences are accepted deliberately and recorded as such: concurrent edits union rather than conflict, and `cat_sort_uniq` sorts lexicographically, so `read_records` must sort by `created_at` on read.
- Repos already `init`'d under #582 hold the broken refspec on disk and must be **migrated**, not merely added-alongside.
- Explicit non-goals: this does not fix #185 (the read-modify-write race within one repo), does not reconcile `NoteRecord.id` collisions, and does not reuse the name `spelunk memory push` (already means "push to the team server").

The implementation (refspec correction + pre-push hook, plus the `init_notes_refspec.rs` test update) lands as one follow-on change once this merges.

## Test plan

Docs-only change; no runtime surface. Verified:
- `docs/adr/069-*.md` is the next free number in the global sequence (checked `docs/adr/` in both spelunk-oss and cloud-api; oss was at 068, cloud-api at 065).
- Every code reference cited in the ADR was confirmed against the tree at this SHA: `configure_notes_refspec` / `FETCH_REFSPEC` (`init.rs:263-264`), `PUSH_HINT` (`init.rs:265`), `read_records` (`git_notes/mod.rs:308`), `hooks.rs` post-commit guard pattern, `tests/init_notes_refspec.rs:27`.
- Repo pre-commit hook ran `cargo fmt --check` + `cargo clippy` clean.

### DIFF
--- a/docs/adr/069-git-notes-sharing-pre-push-hook-and-tracking-refspec.md
+++ b/docs/adr/069-git-notes-sharing-pre-push-hook-and-tracking-refspec.md
@@ -37,8 +37,9 @@ observed behaviour, not projections.
 
 **Publish notes on `git push` through an opt-in, non-blocking pre-push hook;
 merge with `cat_sort_uniq`; fetch into a tracking ref rather than over the live
-one; and have spelunk itself merge that tracking ref on its read paths, so
-reading a teammate's memory needs no opt-in.**
+one; have spelunk itself merge that tracking ref on its read paths, so reading a
+teammate's memory needs no opt-in; and put a lock around the note
+read-modify-write, without which the merge silently eats entries.**
 
 ### D1 – notes sharing is coupled to `git push`, via an opt-in pre-push hook
 
@@ -83,11 +84,22 @@ Two consequences are accepted deliberately:
 
 - **Concurrent edits union rather than conflict.** There is no user
   interaction to resolve, and force-push never enters the picture.
-- **Read order is no longer chronological.** `cat_sort_uniq` sorts lines
+- **Read order is no longer chronological.** Confirmed on merged output, which
+  interleaves lines by `id` rather than by time. `cat_sort_uniq` sorts
   lexicographically, so `read_records`
   (`crates/spelunk-core/src/storage/git_notes/mod.rs`) must sort by `created_at`
   on read. It currently returns records in blob line order, which is
   chronological only because appends happen to land in order today.
+
+**D2's safety rests on git's newline normalization, not on spelunk's code.**
+`append_to_git_notes` builds the body as `format!("{}\n{}", …)`
+(`git_notes/mod.rs:45-49`) with **no trailing newline**. git adds it when
+storing via `notes add -F`. That is the only reason `cat_sort_uniq` does not
+weld the last line of one side onto the first line of the other and corrupt both
+records (verified: every line survives the union parseable). This is load-bearing
+behaviour owned by an external tool, so the implementation must carry a
+**regression test** that would fail if that normalization ever changed. It must
+not be left as an implicit assumption.
 
 ### D3 – opt-in, non-blocking, best-effort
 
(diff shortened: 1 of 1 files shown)


docs/adr/069-git-notes-sharing-pre-push-hook-and-tracking-refspec.md
```diff
@@ -0,0 +1,198 @@
...
+  only via reflog. That is data loss of the product's core asset.
+- **A glob alone is not enough.** `…spelunk*:…spelunk*` fixes the fetch break
+  but still clobbers the local ref. Only the **tracking** destination is safe.
+
+**Consequence, recorded plainly:** fetched notes land in
+`refs/notes/origin/spelunk` and are **not** directly visible to
+`git notes --ref=spelunk` or `spelunk memory list` until merged. "Travels
+automatically on `git fetch`" therefore becomes **fetch + merge**, where the
+merge is D1's hook. Users who already ran `init` under #582 carry the broken
+value on disk, so the implementation must **migrate** that refspec, not merely
```
usercise: No, as discussed we are not worried about this situation. Let's not add a migration for something that almost certainly didn't happen. If anyone built a version of #582 and ran it, they are following enough to read the changelog and resolve manually.

docs/adr/069-git-notes-sharing-pre-push-hook-and-tracking-refspec.md
```diff
@@ -0,0 +1,198 @@
...
+- **The leading `+` on a working ref destroys local notes.** It force-updates
+  the destination: a local unpushed note was silently replaced by the remote's
+  on a plain `git fetch`, reported only as `(forced update)`, and recoverable
+  only via reflog. That is data loss of the product's core asset.
+- **A glob alone is not enough.** `…spelunk*:…spelunk*` fixes the fetch break
+  but still clobbers the local ref. Only the **tracking** destination is safe.
+
+**Consequence, recorded plainly:** fetched notes land in
+`refs/notes/origin/spelunk` and are **not** directly visible to
+`git notes --ref=spelunk` or `spelunk memory list` until merged. "Travels
```
usercise, line 168: Would a complementary fetch hook help here?

docs/adr/069-git-notes-sharing-pre-push-hook-and-tracking-refspec.md
```diff
@@ -0,0 +1,198 @@
...
+`init.rs:265`) tells users to push notes after each memory change, which is
+exactly the orphaning failure in D1; it is replaced by the hook, not kept as a
+fallback.
+
+## Non-goals
+
+- **Not** fixing #185, the read-modify-write race *within* a single repo. That
+  is orthogonal and still open. `cat_sort_uniq` resolves the *inter-repo* race
+  only.
+- **Not** reconciling `NoteRecord.id` collisions. `id` is a local SQLite rowid
```
usercise, line 327: Correct PR [#598](https://github.com/spelunk-cloud/spelunk/pull/598) is in play to resolve this issue outside this ADR.

docs/adr/069-git-notes-sharing-pre-push-hook-and-tracking-refspec.md
```diff
@@ -0,0 +1,198 @@
...
+  and `remote_id` is only set on server sync, so on the local-first path two
+  developers both produce `id:1` for different entries. `cat_sort_uniq` dedupes
+  whole lines, so nothing is lost, but the carrier can legitimately hold
+  colliding ids, and the `init`-time git-notes import must reconcile them.
+- **Not** reusing the name `spelunk memory push`. That already means "push local
+  memory to the team server," and overloading it onto the notes path would
+  conflate two different destinations.
+- **Not** setting `remote.origin.push`, for the reason #582 recorded and this
+  ADR keeps: it would override git's default branch push.
+- **Not** making notes sharing on by default. D3 is opt-in.
```
usercise, line 342: This is fine, as long as we let the user know that there is an additional step for this to work in the output of `spelunk init`.

PR #617, opened 2026-07-14
feat(memory): publish git notes on push via an opt-in pre-push hook

usercise:
Implements ADR-069 **D1 + D3** — the publishing half of git-notes memory sharing. #616 landed reading (the fetch refspec, the read-path merge, the sort); this makes your own memory travel. Opt-in: nothing publishes until you install the hook.

```bash
spelunk hooks install --pre-push
```

From then on every `git push` to a named remote publishes `refs/notes/spelunk` there: fetch the remote's notes onto the tracking ref, merge with `cat_sort_uniq`, push. `hooks uninstall` removes it.

## Why the trigger is `git push`

A note attached to an unpushed commit can reach the remote while the commit does not: origin returns `fatal: could not get object info`, and a fresh clone cannot resolve what the note is attached to. The entry is on the remote and nobody ever sees it.

That makes `git push` the only honest trigger, and it retires `init`'s `PUSH_HINT`, which told users to push notes *after each memory change* and so taught exactly that failure. ADR-069 D4 records it as replaced by the hook, not kept as a fallback. `init`'s summary now announces publishing is opt-in and names the command, so the opt-in is discoverable from the tool's own output rather than only the docs (a D3 requirement on the implementation).

## Four properties are load-bearing, and each is pinned by a test

- **The nested push uses `--no-verify`.** Without it git re-enters this same hook: the observed failure **recursed 740 levels** and stopped only by exhausting the process table (`cannot fork() … Resource temporarily unavailable`), while every outer push still reported **success**. It fails invisibly while pinning the machine. An env sentinel (`SPELUNK_NOTES_PUSH`) is belt and braces on top.
- **The hook never exits non-zero.** A pre-push hook exiting 1 aborts the branch push outright, so a failed publish would cost the user their code push. All six exit paths are `exit 0`, there is no `set -e`, and failures warn on stderr.
- **It never force-pushes.** The push refspec carries no leading `+`. The union already merges both sides, so forcing could only discard a teammate's memory.
- **Only a lost race retries** (bounded at 3): a teammate publishing between our fetch and our push, where re-merging lets the next attempt through. The predicate matches `non-fast-forward` / `fetch first` and nothing else, so an unreachable remote costs one timeout rather than three.

`-s cat_sort_uniq` is passed per invocation, never read from or written to the user's `notes.mergeStrategy` (the default is `manual`, which exits 1 and strands `NOTES_MERGE_WORKTREE`). `remote.origin.push` is never set: any value overrides git's default branch push, the constraint #582 recorded.

Teammates without spelunk are unaffected: the hook skips itself when `spelunk` is not on PATH.

## Test plan

Gates run with the **actual CI commands** (CI uses nextest, not `cargo test`, and lints with `--features rich-formats`), real exit statuses, no pipes:

| Gate | RC |
|---|---|
| `cargo fmt --all -- --check` | 0 |
| `cargo clippy --all-targets --features rich-formats -- -D warnings` | 0 |
| `cargo nextest run` | 0 (1085 passed, 0 leaky) |
| `cargo test --doc` | 0 |
| `cargo nextest run -p spelunk-core --no-default-features` | 0 |
| `cargo test --doc -p spelunk-core --no-default-features` | 0 |

16 tests. Every property was mutation-tested: each was broken deliberately, and the mutation was also run against the *pre-hardening* suite to measure what it would have caught. **Six mutations killed nothing before hardening** — the recursion guard was the only property genuinely under test:

| Mutation | Before | After |
|---|---|---|
| Remove `--no-verify` (sentinel kept) | killed | killed |
| Hook `exit 1` on failed notes push | killed | killed |
| Drop `-s cat_sort_uniq` | killed incidentally | killed directly |
| **Force-push** | **survived** | killed |
| **Retry predicate → match everything** | **survived** | killed |
| **Retry predicate → match nothing** | **survived** | killed |
| Remove foreign-hook bail | killed | killed |
| **Break init install-state detection** | **survived** | killed |
| **Remove `command -v spelunk` bail** | **survived** | killed |
| **Hardcode remote to `origin`** | **survived** | killed |

Two are worth explaining:

**Force-push was unguarded, and the divergence test could not have caught it.** The union merge runs first, so a force-push still carries both sides. The loss only appears when the *fetch* fails and no merge happens. The test breaks only the fetch (`remote.origin.uploadpack`; push rides receive-pack and is unaffected) so the local ref diverges: a normal push is rejected and the teammate's entry survives, `--force` succeeds and destroys it.

**The race test is deterministic, not timed.** A wrapper serves a stale bare clone to the first `upload-pack` and the real origin to the second, gated on a stamp file, so the served *view* changes and nothing depends on timing. Attempt 1 is a genuine `fetch first` rejection; attempt 2 merges and fast-forwards. Widening the predicate is caught by asserting a permanently-rejecting remote is attempted **exactly once**, counted server-side by an `update` hook.

Removing `--no-verify` for the recursion mutation keeps the sentinel, so the nested push re-enters once and stops there: the counter reads 2 instead of 1, enough to kill the test without re-triggering the fork bomb.

## Windows

This is the first test in the repo to execute a git hook, so the `windows-latest` cell has no precedent to lean on. Two Unix-only assumptions were found and fixed by review: the test PATH prepend used a literal `:` (Windows uses `;`, and `:` is also the drive separator, so the binary was never prepended *and* the first real entry was corrupted, silently defeating `command -v spelunk`), and paths embedded in generated shell hooks are now forward-slashed for Git Bash. The hook body itself is POSIX-only and the install writes LF endings.

Verified by reading rather than running, so **watch that cell**: if the fire-count assertion passes but a publish assertion fails, the hook ran and the PATH guard skipped it; if the fire count is 0, git is not invoking the hook at all.

## Known gap, tracked separately

The hook is a shell script and cannot take the in-process notes lock, so its merge runs unlocked and a `spelunk memory add` concurrent with a push can still lose a record. Every in-process caller does take the lock, so the hook is the only unlocked path. Deliberately out of scope here and tracked as a follow-up; the docs carry no caveat for it, since the window is narrow and a warning would outlive the bug.

Refs: ADR-069 (D1, D3, D4).

### DIFF
--- a/crates/spelunk-cli/src/cli/cmd/hooks.rs
+++ b/crates/spelunk-cli/src/cli/cmd/hooks.rs
@@ -9,14 +9,19 @@ pub struct HooksArgs {
 
 #[derive(Subcommand, Debug)]
 pub enum HooksCommand {
-    /// Install a post-commit hook that auto-indexes and harvests memory
+    /// Install a post-commit hook that auto-indexes and harvests memory, or a
+    /// pre-push hook that publishes memory to the remote (`--pre-push`)
     Install(HooksInstallArgs),
-    /// Remove the spelunk post-commit hook
+    /// Remove every git hook spelunk installed
     Uninstall,
 }
 
 #[derive(Args, Debug)]
 pub struct HooksInstallArgs {
+    /// Install the pre-push hook that publishes memory notes on `git push`
+    #[arg(long, conflicts_with = "ci")]
+    pub pre_push: bool,
+
     /// Print a GitHub Actions workflow step instead of writing a git hook
     #[arg(long)]
     pub ci: bool,
@@ -44,6 +49,65 @@ spelunk index "$PROJECT_ROOT" --detach
 spelunk memory harvest --git-range HEAD~1..HEAD --detach
 "#;
 
+const PRE_PUSH_HOOK: &str = r#"#!/bin/sh
+# spelunk pre-push hook — installed by `spelunk hooks install --pre-push`
+# Publishes spelunk memory (refs/notes/spelunk) to the remote you are pushing to,
+# so decisions travel with the code they describe.
+# Silently skips if `spelunk` is not in PATH, so teammates without spelunk are unaffected.
+#
+# This hook never exits non-zero: a failing pre-push hook aborts the branch push
+# outright, and sharing memory must never cost you your push.
+
+if ! command -v spelunk >/dev/null 2>&1; then
+  exit 0
+fi
+
+# Recursion guard. Without it the notes push below re-enters this hook: a version
+# lacking one recursed until the process table was exhausted, while every outer
+# push still reported success. `--no-verify` on that push is the real guard; this
+# sentinel is belt and braces.
+if [ -n "${SPELUNK_NOTES_PUSH:-}" ]; then
+  exit 0
+fi
+
+REMOTE="${1:-origin}"
+
+# Nothing recorded locally, or nothing resolvable to publish to.
+git rev-parse --verify --quiet refs/notes/spelunk >/dev/null 2>&1 || exit 0
+git remote get-url "$REMOTE" >/dev/null 2>&1 || exit 0
+
+attempt=1
+while [ "$attempt" -le 3 ]; do
+  # Onto the tracking ref, never over refs/notes/spelunk: fetching straight onto
+  # the working ref force-updates it and drops notes you have not pushed yet.
+  git fetch --quiet "$REMOTE" "+refs/notes/spelunk:refs/notes/origin/spelunk" >/dev/null 2>&1
+
+  # `-s` is explicit: the notes.mergeStrategy default is `manual`, which exits 1
+  # and strands .git/NOTES_MERGE_WORKTREE. Your own setting is never written.
+  if git rev-parse --verify --quiet refs/notes/origin/spelunk >/dev/null 2>&1; then
+    git notes --ref=spelunk merge -s cat_sort_uniq refs/notes/origin/spelunk >/dev/null 2>&1
+  fi
+
+  # Never forced: the union above already carries both sides, so a force-push
+  # could only discard a teammate's memory.
+  err=$(SPELUNK_NOTES_PUSH=1 git push --no-verify "$REMOTE" \
+    refs/notes/spelunk:refs/notes/spelunk 2>&1) && exit 0
+
+  # Only a lost race is retried: a teammate landed notes between our fetch and
+  # our push, so re-merging theirs lets the next attempt fast-forward. Anything
+  # else (offline, rejected by the remote) would fail identically three times.
+  case "$err" in
+    *non-fast-forward* | *"fetch first"*) ;;
+    *) break ;;
+  esac
+  attempt=$((attempt + 1))
+done
+
+echo "spelunk: could not publish memory notes to '$REMOTE'. Your code push is unaffected." >&2
+echo "spelunk: retry with: git push $REMOTE refs/notes/spelunk" >&2
+exit 0
+"#;
+
 const CI_STEP: &str = r#"# Add to your .github/workflows/ file:
 - name: Update spelunk index
   run: |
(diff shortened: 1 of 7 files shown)


crates/spelunk-cli/src/cli/cmd/hooks.rs
```diff
@@ -44,6 +49,65 @@ spelunk index "$PROJECT_ROOT" --detach
 spelunk memory harvest --git-range HEAD~1..HEAD --detach
 "#;
 
+const PRE_PUSH_HOOK: &str = r#"#!/bin/sh
```
usercise, line 52: This won't work on windows, it isn't guaranteed to have `/bin/sh` to run. 

I think I've changed my mind about the scope of this task. We should create the command spelunk and make the hook a simple spelunk call, not try to script it.

crates/spelunk-cli/src/cli/cmd/hooks.rs
```diff
@@ -44,6 +49,65 @@ spelunk index "$PROJECT_ROOT" --detach
...
+# spelunk pre-push hook — installed by `spelunk hooks install --pre-push`
+# Publishes spelunk memory (refs/notes/spelunk) to the remote you are pushing to,
+# so decisions travel with the code they describe.
+# Silently skips if `spelunk` is not in PATH, so teammates without spelunk are unaffected.
```
usercise, line 56: Hooks doesn't travel with the repo so there is no need to consider if other team mates have spelunk or not. I think it's fair to assume that if you installed spelunk and the hook through it you have spelunk, and that if you remove spelunk that is a valid reason for the hook to fail. It's not the only option for UX, but as a user I'd rather something shouted that there was further clean up required, or that a tool I expected didn't exist. Than for the cruft to just sit around untidied for ever.

crates/spelunk-cli/src/cli/cmd/hooks.rs
usercise: Agreed, and taking the rework. Moving `fetch`/`merge`/`retry` into Rust behind a plumbing command, with the hook reduced to a shim that calls it.

Two notes on the shape:

**The shebang stays, but the surface shrinks from ~200 lines of shell to two.** A git hook has to be an executable file, so the shim is still `#!/bin/sh` + one `exec`. Git for Windows ships its own `sh` and runs hooks through it, which the existing post-commit hook already relies on and the Windows CI cell exercises today. The fragility you're pointing at is the *scripting*, and that's what moves.

**It also closes a gap rather than just relocating one.** The shell hook could not take the in-process notes lock, so its merge ran unlocked and a concurrent `spelunk memory add` during a push could still lose a record (the follow-up filed after the last round). In Rust the publish path takes the same lock as every other writer, so that gap closes as a side effect of this rework instead of needing its own fix.

crates/spelunk-cli/src/cli/cmd/hooks.rs
usercise: You're right that the stated reason is bogus, and the ADR is wrong here too: D3 justifies this guard as "`command -v spelunk` skip so teammates without spelunk are unaffected". `.git/hooks/` is never cloned, so that case cannot happen. Amending the ADR alongside the rework.

One thing worth flagging before it goes, because the guard turns out to be load-bearing for a reason nobody wrote down: `install.sh` falls back to `~/.local/bin` when `/usr/local/bin` isn't writable, and tells you to add it to your **shell profile**. macOS GUI apps get their PATH from launchd, not your profile, so Tower / GitHub Desktop / VS Code / IntelliJ run hooks without `~/.local/bin` on PATH. Dropping the guard as written would break `git push` from every GUI client for anyone on that install path, which is a different failure from the one you want to shout about.

Resolution (founder call): **`hooks install` embeds the absolute path of the binary it resolved**, instead of relying on PATH. That separates the two cases cleanly, which `command -v` never could:

- **spelunk genuinely removed** -> the absolute path fails, the hook exits non-zero, the push stops and says so. Exactly the shout you want, no cruft sitting around untidied.
- **spelunk present but not on a GUI client's PATH** -> irrelevant, since there is no PATH lookup.
- **publish fails** (offline, remote rejects) -> still exits 0, so memory sharing can never cost someone their code push.

Cost, recorded honestly: the embedded path goes stale if you move or reinstall the binary. It fails loudly and `spelunk hooks install --pre-push` re-resolves it, which is the tradeoff you picked over silence.

PR #633, opened 2026-07-16
docs(adr): ADR-071: per-server scoping of the client bearer credential

usercise:
## Summary

Records the decision for scoping the client's team-server credential per server, replacing the single flat `server_key` that cannot serve ADR-056's separate-instances topology.

- **D1**: one secret-store entry (`service=spelunk, user=server_keys`) holding a JSON origin→key map (single keychain item by design: per-item macOS prompts).
- **D2**: per-`server_url` resolution precedence: `SPELUNK_SERVER_KEY` env → `[auth]` token → `server_keys[origin]` → legacy flat entry (deprecation nudge only when a map exists), as `Config::bearer_for(server_url)`.
- **D3**: new `spelunk auth set-key --server <url>` (stdin/prompt, never argv) and `auth list-servers`; `logout` clears map + legacy entry. Also corrects `save_server_key`'s stale doc comment (claims `spelunk login` persists it; nothing does).
- **D4**: `server_key` removed from committed project config (warn-and-ignore with a pointer at `auth set-key`) — pre-v1.0 breaking window, `memory_server_*` removal as precedent.

Non-goals fence off settled ground: ADR-056 tenancy (single trust domain per instance) and ADR-066 transport (native TLS, no reverse proxy) are unchanged.

**Needs founder review before any implementation is dispatched** — notably D4 (removing a documented config field) is the founder-visible call.

## Verification
- Every cited file:line checked against `main` at authoring time (precedence at `config.rs:573-591`; zero production callers of `save_server_key`; the plaintext-key doc instruction at `server.md:159-164`; `ProjectConfig` accepting `server_key` at `config.rs:256-266`).
- ADR number 071 verified free across both repos' `docs/adr/` sequences.
- Doc-only diff: no code changes; `cargo fmt`/clippy pre-commit hook passed on commit.

### DIFF
--- a/docs/adr/071-per-server-client-bearer-scoping.md
+++ b/docs/adr/071-per-server-client-bearer-scoping.md
@@ -75,9 +75,12 @@ the credential a real home, keyed by the server it belongs to.
 ## Decision
 
 **Store a per-origin key map in the secret store as a single entry; resolve
-the bearer for a given `server_url` through it; give the key a real command
-surface (`spelunk auth set-key`, `spelunk auth list-servers`); and remove
-`server_key` from the committed project config.**
+the bearer for a given `server_url` through it by branching on credential kind
+rather than probing every tier; give the key a real command surface
+(`spelunk auth set-key`, `spelunk auth list-servers`, and a scoped
+`spelunk logout --servers` / `--server <url>`); remove `server_key` entirely
+from the committed project config; and migrate the legacy flat key into the
+map instead of reading both indefinitely.**
 
 ### D1 – one secret-store entry holding a per-origin key map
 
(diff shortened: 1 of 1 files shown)


docs/adr/071-per-server-client-bearer-scoping.md
```diff
@@ -0,0 +1,282 @@
...
+  key is read from stdin or an interactive prompt, **never** from argv: a
+  positional or flag-valued secret lands in shell history and in `ps` output,
+  which is the same class of leak D4 closes for the committed file. The URL
+  is normalized to its origin before storage, so `set-key` and resolution
+  cannot disagree about spelling.
+- **`spelunk auth list-servers`** prints the origins present in the map, and
+  whether a legacy flat key also exists. It never prints key material, not
+  even truncated: a listing surface that shows secret prefixes trains users
+  to have secrets on screen.
+- **`spelunk logout`** (existing) additionally clears the map. It already
```
usercise: This should be a flag or an interactive prompt. Just because you are trying to an issue with cloud access shouldn't mean that you delete your access keys automatically.

docs/adr/071-per-server-client-bearer-scoping.md
```diff
@@ -0,0 +1,282 @@
...
+will actually go to, with this precedence (highest first):
+
+1. **`SPELUNK_SERVER_KEY` environment variable.** Unchanged, and still a
+   per-invocation override: CI, headless setups, and one-off testing against
+   a server whose key is not stored keep working exactly as today.
+2. **`[auth].access_token`.** Unchanged. The cloud login path is a different
+   credential kind with its own refresh lifecycle, and this ADR does not
+   touch it.
+3. **`server_keys[origin]`**, the map from D1, looked up by the normalized
+   origin of the resolved `server_url`.
```
usercise: Wouldn't this mean we're back to two keychain prompts? I'm not saying we can't keep this as is. I'm just saying we need to take that into account, and make sure we only prompt for the right key when needed.

docs/adr/071-per-server-client-bearer-scoping.md
```diff
@@ -0,0 +1,282 @@
...
+4. **The legacy flat secret-store `server_key` entry**, as a back-compat
+   fallback. When this tier answers *and* a non-empty map exists, a one-line
+   deprecation nudge on stderr points at `spelunk auth set-key`; when no map
+   exists (the single-server user who has never touched the new surface) it
+   answers silently, because for that user nothing is wrong.
+
+This lands as a `Config::bearer_for(server_url)` lookup rather than a field
```
usercise, line 168: Do we need to add a Config(use_cloud) or similar so that we don't have to infer if we need a cloud auth or a server api key?

docs/adr/071-per-server-client-bearer-scoping.md
```diff
@@ -0,0 +1,282 @@
...
+persistence path, which until now existed only for its tests, and it retires
+the plaintext-file edit as the documented way to install a key. The stale doc
+comment on `save_server_key` (claiming `spelunk login` persists it) is
+corrected as part of this work.
+
+### D4 – `server_key` is removed from the committed project config
+
+`ProjectConfig` stops accepting `server_key`. A `.spelunk/config.toml` that
+still carries one gets a load-time warning naming the file, stating that the
+value is ignored, and pointing at `spelunk auth set-key --server <url>` as
```
usercise: The warning should also tell the user to cycle the key as it would now have been exposed in a public forum.

docs/adr/071-per-server-client-bearer-scoping.md
```diff
@@ -0,0 +1,282 @@
...
+them fix their own machine with `auth set-key` immediately and the repo at
+leisure. The other shareable fields (`server_url`, `project_id`,
+`server_ca`) are unaffected; they are what the committed file is for.
+
+### Migration and back-compat
+
+- **The legacy flat entry keeps working** as resolution tier 4 (D2). A
+  single-server user who never runs `auth set-key` sees zero behavior change:
+  same key, same store, same precedence position relative to env and
+  `[auth]`.
```
usercise: Again this leads to additional keychain prompts does it not? I would prefer if we migrate it, we can deprecate the migration at create time, letting it be known it will be removed in version X. That gives people an upgrade path if they need it but doesn't burden us with dead code in the long run.

docs/adr/071-per-server-client-bearer-scoping.md
```diff
@@ -0,0 +1,282 @@
...
+- **The existing plaintext-file migration (`config.rs:524-541`) is
+  unchanged.** It still moves a bare `server_key` out of the personal
+  `config.toml` into the flat store entry. It does not learn about the map:
+  a migrated flat key serves tier 4, and the user graduates to the map only
+  by running `auth set-key`. Teaching the migration to guess an origin for a
+  bare key would require guessing which server the key belongs to, which is
+  exactly the ambiguity the map exists to remove.
```
usercise: This code path should just be removed. If someone hasn't had the plaintext-file migration trigger yet they must be in the onboarding journey. They would just need to read the docs again. I'd rather we just don't read server_key from any config.toml. Otherwise we risk mixing messages and someone in desperation may copy or accidental create the ~/.local/spelunk/config.toml to project/.spelunk/config.toml.

docs/adr/071-per-server-client-bearer-scoping.md
```diff
@@ -0,0 +1,282 @@
...
+- **No data migration is needed** because no released surface writes the map
+  today; it is born empty and populated only by `auth set-key`.
+- **Docs follow, separately.** The client-configuration sections of
```
usercise: No let's update them at the same time. Why would we release a version that tightens security and not update the docs with those new improvements?

PR #634, opened 2026-07-16
release: build Linux artifacts against glibc 2.31 in debian:11 container

usercise:
## Summary

Linux release artifacts were silently incompatible with Debian 11 and Ubuntu 20.04 (glibc 2.31 floor) because the build runner (ubuntu-24.04) uses glibc 2.39. Binaries linked against newer symbols immediately crash on install.

This PR fixes it by building x86_64 and aarch64 Linux targets inside a debian:11 (Bullseye) container, ensuring all linked symbols belong to glibc <= 2.31. A new CI step (`Enforce glibc ceiling`) runs objdump on both binaries and fails the release if either symbol exceeds 2.31. The check has a malformed-symbol fail-safe: empty, non-numeric, or GLIBC_PRIVATE results are treated as broken and cause exit 1.

Smoke tests now run on 3 images (debian:11, ubuntu:20.04, ubuntu:24.04 for .deb) and 4 images (debian:11, ubuntu:20.04, ubuntu:24.04, debian:12 for tarballs), exercising the full CLI recipe (git init, memory add, memory list) so install-success + binary-runs are proven together. Both smoke jobs are now dependencies of the release job, making them load-bearing for publish.

The .deb package declares libdbus-1-3 as a runtime dependency (auto-derived via dpkg-shlibdeps). Tarball users on minimal images must install libdbus-1-3 manually.

### Local execution (engineer and test stages)
1. Built debian:11 container on the commit and ran the ceiling check against spike binaries:
   - x86_64 binary: max versioned glibc symbol = GLIBC_2.31
   - aarch64 binary: max versioned glibc symbol = GLIBC_2.31
2. Extracted tarballs and ran the smoke test scripts verbatim on each image:
   - debian:11: spelunk status, init, memory add, memory list all pass
   - ubuntu:20.04: spelunk status, init, memory add, memory list all pass
   - ubuntu:24.04: spelunk status, init, memory add, memory list all pass
3. Ran ceiling check with malformed-symbol scenarios and verified fail-safe exits 1

### Actions validation
The container-job path itself (the `runs-on: ubuntu-latest` job with `container: debian:11`) is first exercised on the next RC tag. The smoke jobs themselves (which run in ubuntu/debian containers on separate runner VMs) have been tested locally with the same workflow commands and images shown above.

aarch64 native runner (ubuntu-24.04-arm) was substituted with ubuntu-latest in test stage for ceiling check verification; the actual release run will use ubuntu-24.04-arm per matrix config.

No ADR required; floor is inferred from the artifacts and enforced by CI.


.github/workflows/release.yml
```diff
@@ -49,78 +86,64 @@ jobs:
             ~/.cargo/registry
             ~/.cargo/git
             target
-          key: ${{ runner.os }}-cargo-${{ matrix.target }}-${{ hashFiles('**/Cargo.lock') }}
+          # The `-bullseye2.31-` component keeps container-built artifacts
+          # separate from caches produced by builds on the runner VM. cc-built
+          # static archives (ring, bundled sqlite, blake3) do not refingerprint
+          # on the C toolchain's glibc, so restoring a VM-era cache into the
+          # container would silently re-raise the binaries' glibc ceiling.
+          key: ${{ runner.os }}-bullseye2.31-cargo-${{ matrix.target }}-${{ hashFiles('**/Cargo.lock') }}
```
usercise:
```suggestion
          key: ${{ runner.os }}-${{ matrix.container }}-${{ matrix.target }}-${{ hashFiles('**/Cargo.lock') }}
```

PR #669, opened 2026-07-18
docs: consolidate server-setup, split third-party-models

usercise:
## Summary

Four related docs/hygiene fixes, batched into one PR because they touch overlapping sections of the same consolidated doc.

- **spelunk-oss^224** — Consolidate `docs/server.md` + `docs/self-hosting.md` into a single `docs/server-setup.md` (naming aligned with the marketing site's `server-setup` slug). API reference moved to `docs/architecture/server-api.md` (also corrected for drift against shipped code: health JSON fields, `/index/embed` response content-type). Local-auto-server material reduced to a stub pointing at `getting-started.md`. Old filenames left as redirect stubs. Fixes the self-contradicting "each developer adds a `.spelunk/config.toml`... (commit it)" sentence, and restructures the Docker team-server recipe so `docker compose --profile team-server` is the primary path (bare `docker run` as a secondary manual alternative), drops `--rm` from the long-lived server container, and adds a `docker logs`/`docker compose logs` troubleshooting note.
- **spelunk-oss^227** — Team-server TLS docs never minted a cert. Adds a verified two-step self-signed/internal-CA recipe (CA mint -> leaf CSR with `subjectAltName`, `basicConstraints=CA:FALSE`, `extendedKeyUsage=serverAuth`) to `docs/server-setup.md`, documents the SAN-only (no CN fallback) rustls requirement, and adds the missing `docker build` step to the Docker walkthrough.
- **spelunk-oss^223** — `docs/third-party-models.md` was pure license attribution with no configuration content. Split attribution into `docs/model-attribution.md`; rewrote `third-party-models.md` to document the external LLM (`--llm-url`) and embedding endpoint configuration story, cross-linked from `server-setup.md`.
- **spelunk-oss^226** — `sync` was documented as an alias of `memory push` but is a distinct two-way command. Extracted a shared `server_url` guard (`crates/spelunk-cli/src/cli/cmd/memory/{sync,push}.rs`) with one actionable message, and fixed `CLAUDE.md`'s Common Commands to describe the real one-way (`push`) vs two-way (`sync`) relationship.

## Test plan

- [x] `cargo fmt --all -- --check` clean (isolated `CARGO_TARGET_DIR`)
- [x] Full gate green (fmt/clippy/tests, both feature configs, doc-tests) in an isolated target dir
- [x] TLS cert recipe verified end-to-end against real `spelunk-server`/`spelunk` binaries: broken (`CA:TRUE`) recipe reproduces `invalid peer certificate: Other(OtherError(CaUsedAsEndEntity))`; corrected recipe reaches `Capability tier: Server`
- [x] `docker compose --profile team-server up -d` invocation verified against `docker-compose.yml`'s real `team-server` profile (env vars, `restart: unless-stopped`, volume mounts)
- [x] Redirect stubs (`docs/server.md`, `docs/self-hosting.md`) resolve; all new cross-link anchors checked against real headings in `docs/server-setup.md`
- [x] Branch fetched and confirmed even with `origin/main` (no rebase needed)


crates/spelunk-cli/src/capability.rs
```diff
@@ -908,6 +908,30 @@ pub fn require_tier1(feature: &str, tier: &Tier, server_url: Option<&str>) -> an
...
+///
+/// This is deliberately **explicit-config-only**: it reads `cfg.server_url`
+/// and nothing else, and never probes reachability (so it produces a message
+/// before any network call is made, and stays correct even offline). Returns
+/// the configured `server_url` on success.
+pub fn require_team_server(feature: &str, cfg: &Config) -> anyhow::Result<String> {
+    cfg.server_url.clone().ok_or_else(|| {
+        anyhow::anyhow!(
+            "'spelunk {feature}' requires a server. Set `server_url` in your spelunk config \
+             (e.g. ~/.config/spelunk/config.toml or .spelunk/config.toml)."
```
usercise, line 933: I'm not sure "~/.config/spelunk/config.toml" can act as source for server_url. Even if it can I'm not convinced it should. If we are setting up a team server for a project I would want all the users to use the same team server so I'd like that checked in with the code.

crates/spelunk-cli/src/cli/cmd/memory/push.rs
usercise, line 1: This PR is marked as docs only but we're changing the logic in it. Please split that this out to it's own PR so it can be reviewed properly.

crates/spelunk-cli/src/capability.rs
```diff
@@ -908,6 +908,30 @@ pub fn require_tier1(feature: &str, tier: &Tier, server_url: Option<&str>) -> an
...
+/// auto-discovered loopback inference server makes the tier `Server` while
+/// `cfg.server_url` stays `None`, and that loopback server is never a memory
+/// store (ADR-004). Callers therefore check `require_tier1` for reachability
+/// first, then this guard for explicit team-server configuration.
+///
+/// This is deliberately **explicit-config-only**: it reads `cfg.server_url`
+/// and nothing else, and never probes reachability (so it produces a message
+/// before any network call is made, and stays correct even offline). Returns
+/// the configured `server_url` on success.
+pub fn require_team_server(feature: &str, cfg: &Config) -> anyhow::Result<String> {
```
usercise: Not the right name, this will work equally well with cloud server as it will with team server.

crates/spelunk-cli/src/capability.rs
```diff
@@ -908,6 +908,30 @@ pub fn require_tier1(feature: &str, tier: &Tier, server_url: Option<&str>) -> an
...
+///
+/// `require_tier1` alone is not sufficient for these commands: an
+/// auto-discovered loopback inference server makes the tier `Server` while
+/// `cfg.server_url` stays `None`, and that loopback server is never a memory
+/// store (ADR-004). Callers therefore check `require_tier1` for reachability
+/// first, then this guard for explicit team-server configuration.
+///
+/// This is deliberately **explicit-config-only**: it reads `cfg.server_url`
+/// and nothing else, and never probes reachability (so it produces a message
+/// before any network call is made, and stays correct even offline). Returns
```
usercise: Working offline makes sense I guess, but what is the point of us returning capabilities from the server health endpoint if the client doesn't use those capabilities. Also since this call happens on a single run cli execution why return server_url if we know we are offline? It's not like we will be able to make a call in the nano-seconds from now when the server_url will be used.

crates/spelunk-cli/src/capability.rs
usercise, line 1: This is file is too long and don't pass our file size rules.

docs/architecture/server-api.md
```diff
@@ -1,22 +1,22 @@
...
-OAuth2 in future, and the changes required to existing endpoints.
+`spelunk-server`: the `AuthProvider` trait, and every route the server exposes.
 
-The server already has a working memory API (`src/server/handlers.rs`). This
-document covers:
-
-1. The `AuthProvider` trait (replaces the inline `auth_middleware` function).
-2. Changes to existing endpoints.
-3. New endpoints needed for CLI integration.
+1. The `AuthProvider` trait (replaced the original inline `auth_middleware` function).
```
usercise: I don't care about the history here. Why should we care that v0.2 used `auth_middleware`?

docs/architecture/server-api.md
```diff
@@ -50,14 +50,15 @@ DB table is empty after `/v1/projects/{id}/index/embed` returns, and that
 
 ## Auth architecture
 
-### Current state (to be refactored)
+### History
```
usercise: Why?

docs/architecture/server-api.md
```diff
@@ -159,28 +160,19 @@ let auth: Arc<dyn AuthProvider> = Arc::new(ApiKeyAuth::from_env());
...
-inline `auth_middleware` in `lib.rs` is replaced by the trait-driven version
-shown above; `state.api_key: Option<String>` is removed in the same change.
-No handler files (`handlers.rs`) require edits beyond switching consumers from
-`api_key` reads to `Extension<AuthContext>` extraction where they care about
-the principal (today, none do).
+The trait and `ApiKeyAuth` impl above are implemented verbatim in
+`crates/spelunk-server/src/auth.rs`, wired in via `pub mod auth;` in
+`crates/spelunk-server/src/lib.rs`. `state.api_key: Option<String>` no longer
+exists. No handler in `handlers.rs` reads the principal today, so none extract
+`Extension<AuthContext>`.
```
usercise: Again why do we care about the history, and why does the architecture docs document individual function signatures, that is not the right granularity it will just lead to docs that are out of date and sync.

docs/self-hosting.md
usercise, line 1: Just delete the file

docs/server.md
usercise, line 1: Just delete the file

PR #753, opened 2026-07-24
Paginate memory sync/pull past the server's 100-entry page limit

usercise:
## Problem

`spelunk sync` and `spelunk memory pull` share a single pull path (`pull_and_apply_since`) that called `CloudSyncClient::pull_since` exactly once and treated whatever it returned as the entire backlog. `pull_since` never sent an explicit `limit` query param, so the server silently applied its own 100-entry default (`SinceQuery`'s `default_since_limit()` in `handlers.rs`). A first sync into an established project with more than 100 pending entries reported success and printed an "Applied N" count, but only the first page ever landed locally — no error, no indication anything was missing.

## Fix

- `spelunk-core`: `pull_since` now sends an explicit `limit=500` (`CloudSyncClient::PULL_PAGE_LIMIT`, matching the server's own clamp ceiling) on every request. It still fetches exactly one page — no looping at this layer.
- `spelunk-cli`: `pull_and_apply_since` now loops — applies each page as it arrives, advances the cursor to the last entry's `id`, and keeps going until a page comes back shorter than the requested limit (the definitive "nothing left" signal, including the empty-page case). Since both `spelunk sync`'s `sync_round` (both pull passes) and `spelunk memory pull`'s `pull_and_apply` call through this same function, the fix covers both commands in one place. The completion message needed no wording change — `pulled` already reflects every page once the loop is exhaustive.

Looping (not just raising the limit) was the deliberate choice: the server clamps `limit` to `1..=500`, so a fixed higher limit alone reproduces the exact same silent-truncation bug once any project's backlog exceeds 500 pending entries.

## Test plan

Ran independently in the review worktree, not just trusting prior stages' reported numbers:

- `SPELUNK_SECRET_STORE=file cargo fmt --all -- --check` — clean.
- `SPELUNK_SECRET_STORE=file cargo clippy --all-targets --features rich-formats -- -D warnings` — clean, 0 warnings (the real CI invocation).
- `SPELUNK_SECRET_STORE=file cargo clippy --workspace --all-targets --no-default-features -- -D warnings` — clean, 0 warnings.
- `SPELUNK_SECRET_STORE=file cargo test -p spelunk-cli -p spelunk-core` — full suite green, 0 failed (checked the bare exit code directly, not piped through `tail`, since zsh has no `PIPESTATUS`). 501 passed in the largest suite; every other suite in both crates green; no flakes this run.
- Read the full diff against `origin/main` line by line against the architect's 10-item acceptance test list (task description v2) — all 10 items have a corresponding test, plus 4 additional hardening tests from test-engineer (mid-pagination failure durability/retry, malformed-entry atomicity, ignoring a lying `count` field, and a spelunk-core-level 500-vs-404 distinction test). Verified the new tests assert real request counts and store state via wiremock exact-match query params, not pasted/tautological expected values.
- Grepped the full diff for internal board/task-id leakage (`spelunk-oss#`, `^281`, `^282`) — none found.
- Checked the new CHANGELOG.md entry: sits under `[Unreleased] ### Fixed`, accurately describes the user-visible fix, no em-dashes. (New Rust doc comments in this diff do use em-dashes, consistent with this file's — and the wider codebase's — pre-existing, pervasive em-dash convention in source comments; not flagged as this diff introducing an inconsistency.)
- No `Cargo.toml`/`Cargo.lock` changes, so no new dependencies to audit.

## Test list coverage (architect's 10 items, task spelunk-oss^282)

1. Single page (≤ one page) — unchanged single-request behavior.
2. Two pages (500 + 150) — cursor advances to prior page's last id, no duplicates.
3. Three-plus pages (500 + 500 + 200) — loop continues past two iterations.
4. Full page always triggers exactly one more request.
5. Already fully synced (empty first page) — terminates after one request.
6. `spelunk sync`'s completion message reports the true multi-page total.
7. `spelunk memory pull` (one-way) also paginates fully.
8. `sync_round`'s second (confirmation) pull pass paginates fully independently.
9. Idempotency across pagination / re-runs — no double-counting.
10. Existing single-page regressions preserved (404-is-empty, `None` cursor starts at nil UUID).


crates/spelunk-cli/src/cli/cmd/memory/sync.rs
```diff
@@ -2345,6 +2367,338 @@ mod tests {
         assert_eq!(pulled_again, 0);
     }
 
+    // ── pull pagination: exhaust every page, not just the first ─────────────
```
usercise: This file is much too long, needs splitting up.

crates/spelunk-core/src/storage/remote/sync.rs
```diff
@@ -312,11 +312,32 @@ impl CloudSyncClient {
...
+    /// [`pull_since`](Self::pull_since) fetches exactly one page: it never
+    /// requests more than this many entries and never loops. A caller that
+    /// needs a store's full backlog (see `pull_and_apply_since` in
+    /// spelunk-cli) must keep calling with the cursor advanced to the last
+    /// entry's `id` until a page comes back shorter than
+    /// [`PULL_PAGE_LIMIT`](Self::PULL_PAGE_LIMIT) — the definitive "nothing
+    /// left" signal, including the empty-page case. A response exactly at
+    /// this limit does not by itself prove more remain, but it can never be
+    /// treated as the last page either, since the server never returns more
+    /// than this many even when more exist.
```
usercise: Why are we documenting `pull_since` here? Just document the const.

crates/spelunk-core/src/storage/remote/sync.rs
```diff
@@ -312,11 +312,32 @@ impl CloudSyncClient {
         anyhow::bail!("DELETE /memory/{remote_id} failed ({status}): {text}")
     }
 
-    /// Pull entries after the UUIDv7 cursor `since_id` (the max cloud
-    /// `remote_id` already synced locally — decision #183). When `since_id` is
-    /// `None` (nothing synced yet) this is a full catch-up: the
-    /// nil UUID `00000000-…` sorts before every UUIDv7, so it returns all
-    /// entries.
+    /// Entries requested per `/memory/since` page. Matches the server's own
```
usercise: Is this true, is this const only valid/used for memory since? If so it should be renamed to not confuse it with a generic page limit.

crates/spelunk-core/src/storage/remote/sync.rs
```diff
@@ -312,11 +312,32 @@ impl CloudSyncClient {
...
+    /// clamp ceiling (`ServerDb::notes_since_id`'s `limit.clamp(1, 500)`), so
+    /// a single call already asks for the most the server will ever return
+    /// in one response, minimizing round trips.
```
usercise: This should only talk about the maximum the server accept. Just because server accept that doesn't mean our api calls need to use `PULL_PAGE_LIMIT`, only that `limit <= PULL_PAGE_LIMIT`.

crates/spelunk-core/src/storage/remote/sync.rs
```diff
@@ -535,6 +557,7 @@ mod tests {
                 "since_id",
                 "01890000-0000-7000-8000-000000000000",
             ))
+            .and(query_param("limit", "500"))
```
usercise: Why isn't this using the const? We test the const as a unit below, so that can't be the reason.

crates/spelunk-core/src/storage/remote/sync.rs
```diff
@@ -328,9 +349,10 @@ impl CloudSyncClient {
         // The all-zero UUID precedes every UUIDv7 in `id > $cursor` order, so it
         // is the natural "from the beginning" cursor for a first sync.
         let cursor = since_id.unwrap_or("00000000-0000-0000-0000-000000000000");
+        let limit = Self::PULL_PAGE_LIMIT.to_string();
```
usercise: Why did we change limit from 100? That was the default server side, which is what we got without a query param. That still sounds like a good default, 500 may be too big for slow network or with large memory entries. The important thing was that we set a limit so we could do pagination, as the api doesn't have pagination data.

PR #755, opened 2026-07-24
Make the test suite hermetic-by-construction against ambient git config

usercise:
## Summary

The test suite already isolated itself from a developer's ambient global/system
git config via `GIT_CONFIG_GLOBAL`/`GIT_CONFIG_SYSTEM=/dev/null` env-var
redirection, but that mechanism was applied by convention only: a manually
invoked, four-times-copy-pasted `isolate_git_config()` helper with nothing
stopping a new test file from spawning `git` and forgetting to call it.
`crates/spelunk-core/tests/worktree_index_resolution.rs` had already done
exactly that, despite its own doc-comment claiming to be "fully hermetic."
Separately, one test's mid-test global-env patch-then-restore wasn't
panic-safe.

This closes those gaps:

- **Structural CI guard.** New `scripts/check-git-isolation.sh` fails the
  build if a test file spawns `git` (`Command::new("git")`, however wrapped,
  whitespaced, or aliased) without wiring in the isolation fixture. Wired in
  as the first step of CI's Check & Lint job (no cargo dependency, fails
  fast). Self-tests itself against 7 fixtures (`--self-test`).
- **`worktree_index_resolution.rs` fixed.** Now routes through a new
  `common::git_command(cwd)` constructor in `spelunk-core`'s
  `tests/common/mod.rs` that always calls `isolate_git_config()` first,
  making the "fully hermetic" doc-comment true rather than aspirational.
  Backed by a regression test that re-execs the binary under a hostile
  ambient `core.hooksPath` and confirms isolation shadows it.
- **Identity env vars also cleared.** `GIT_CONFIG_GLOBAL`/`GIT_CONFIG_SYSTEM`
  only redirect config *files*; they don't touch
  `GIT_AUTHOR_*`/`GIT_COMMITTER_*`/`EMAIL`, which git resolves before
  consulting config at all. All four `isolate_git_config()` copies now clear
  those too, backed by a re-exec regression test proving a poisoned ambient
  identity env doesn't leak into a test's commits.
- **Panic-safe restore.** `ensure_notes_rewrite_ref_honours_a_users_global_value`'s
  manual set/restore-before-assert (which only protected against a panic in
  the assertions, not the awaited call itself) is replaced with an RAII
  `RestoreGitConfigGlobal` guard whose `Drop` restores unconditionally,
  including on unwind. Backed by a `catch_unwind`-based test proving the
  guard restores after a forced panic.
- **Consolidated.** `integration_git_notes.rs`'s private `isolate_git_config`
  moved into `spelunk-core`'s shared `tests/common/mod.rs`, so
  `worktree_index_resolution.rs`'s fix could reuse it instead of copy-pasting
  a third time. The remaining duplication (4 physical definitions, one per
  {crate, src/tests-boundary-side}) is the floor Rust's compilation
  boundaries allow — each is now the sole canonical copy for its pairing,
  not an independent copy-paste.
- **Docs synced.** `docs/testing.md`'s "Ambient git config in tests" section
  updated: the identity-var clearing, the moved call site, and the
  previously-true-but-now-false claim that "CI runners... a missing call
  here never fails CI" (the whole point of this PR is that it now does).
- **No CHANGELOG entry.** Pure test/CI-infrastructure change with no
  user-facing behavior change, consistent with this repo's existing
  precedent (CHANGELOG's only comparable prior entry, `0.8.3`'s
  `### Internal`, is the exception rather than the rule; most equivalent
  test/CI-only changes get none).

## Test plan

All commands run with `SPELUNK_SECRET_STORE=file` in the PR branch's worktree
(fresh/cold `target/`, so full recompiles):

- `cargo test --workspace` — exit 0. 70 "test result: ok" blocks across the
  workspace (`spelunk-core`, `spelunk-cli`, `spelunk-server`, doctests
  included), 0 FAILED.
- `cargo clippy --all-targets --features rich-formats -- -D warnings` — exit 0.
- `cargo clippy --workspace --all-targets --no-default-features -- -D warnings` — exit 0.
- `cargo fmt --all -- --check` — exit 0, no diff.
- `./scripts/check-git-isolation.sh --self-test` — `OK` (7 fixtures: good,
  unrelated, 4 "bad" spawn-shapes including multiline/whitespace/aliased
  `Command::new`, and a src/-side `#[cfg(test)]`-only violation).
- `./scripts/check-git-isolation.sh .` against the current (fixed) tree —
  `OK`, zero violations.
- Manually confirmed the CI workflow step (`.github/workflows/ci.yml`) runs
  the lint as the very first step of the Check & Lint job, before any cargo
  invocation.

Additionally reviewed:
- Full diff read against `origin/main` and checked against every acceptance
  criterion in the task description — satisfied.
- Grepped the full diff for internal board/task-id references and em-dashes
  — none found.
- Verified the docs-writer's "no CHANGELOG entry" call by checking
  `CHANGELOG.md` directly for precedent (`0.8.3`'s lone `### Internal`
  section) — confirms the claim.

### DIFF
--- a/crates/spelunk-cli/tests/plumbing_helpers.rs
+++ b/crates/spelunk-cli/tests/plumbing_helpers.rs
@@ -21,6 +21,12 @@ use tempfile::TempDir;
 ///
 /// `/dev/null` is not a Windows path, but git skips a scope whenever its var
 /// is set, whatever the path resolves to, so this isolates on Windows too.
+///
+/// `GIT_CONFIG_GLOBAL`/`GIT_CONFIG_SYSTEM` only redirect config *files*;
+/// they don't touch `GIT_AUTHOR_*`/`GIT_COMMITTER_*`/`EMAIL`, which git
+/// consults before config and so override this file's fabricated
+/// `user.name`/`user.email` if the ambient process (a developer's shell, a
+/// CI runner's bot identity) happens to export them. Those are cleared too.
 pub fn isolate_git_config() {
     static ONCE: std::sync::Once = std::sync::Once::new();
     ONCE.call_once(|| {
@@ -30,6 +36,17 @@ pub fn isolate_git_config() {
         unsafe {
             std::env::set_var("GIT_CONFIG_GLOBAL", "/dev/null");
             std::env::set_var("GIT_CONFIG_SYSTEM", "/dev/null");
+            for var in [
+                "GIT_AUTHOR_NAME",
+                "GIT_AUTHOR_EMAIL",
+                "GIT_AUTHOR_DATE",
+                "GIT_COMMITTER_NAME",
+                "GIT_COMMITTER_EMAIL",
+                "GIT_COMMITTER_DATE",
+                "EMAIL",
+            ] {
+                std::env::remove_var(var);
+            }
         }
     });
 }
--- a/crates/spelunk-cli/tests/plumbing_helpers.rs
+++ b/crates/spelunk-cli/tests/plumbing_helpers.rs
@@ -3,53 +3,18 @@
 //! Every test that needs an indexed project DB should call
 //! `index_fixture_project()`.  Tests that need no index still share helpers
 //! for constructing `Command` instances.
-#![allow(dead_code)]
+#![allow(dead_code, unused_imports)]
 
 use assert_cmd::Command;
 use std::path::{Path, PathBuf};
 use tempfile::TempDir;
 
-/// Drop the machine's global/system git config for every git this process
-/// spawns, including a setup git a test runs directly. Must be process-wide,
-/// not per-`Command`: a helper only controls the git it spawns itself, never
-/// one the code under test spawns for itself.
-///
-/// A temp repo's local config does not shadow an ambient value the repo never
-/// sets. A global `commit.gpgsign = true` makes setup commits sign as the
-/// fabricated test identity below, which no contributor holds a key for, so
-/// the commit exits non-zero.
-///
-/// `/dev/null` is not a Windows path, but git skips a scope whenever its var
-/// is set, whatever the path resolves to, so this isolates on Windows too.
-///
-/// `GIT_CONFIG_GLOBAL`/`GIT_CONFIG_SYSTEM` only redirect config *files*;
-/// they don't touch `GIT_AUTHOR_*`/`GIT_COMMITTER_*`/`EMAIL`, which git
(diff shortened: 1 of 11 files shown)


crates/spelunk-cli/tests/plumbing_helpers.rs
```diff
@@ -21,6 +21,12 @@ use tempfile::TempDir;
 ///
 /// `/dev/null` is not a Windows path, but git skips a scope whenever its var
 /// is set, whatever the path resolves to, so this isolates on Windows too.
+///
+/// `GIT_CONFIG_GLOBAL`/`GIT_CONFIG_SYSTEM` only redirect config *files*;
+/// they don't touch `GIT_AUTHOR_*`/`GIT_COMMITTER_*`/`EMAIL`, which git
+/// consults before config and so override this file's fabricated
+/// `user.name`/`user.email` if the ambient process (a developer's shell, a
+/// CI runner's bot identity) happens to export them. Those are cleared too.
 pub fn isolate_git_config() {
```
usercise: Why do we need multiple versions of this same function?

crates/spelunk-core/tests/common/mod.rs
```diff
@@ -34,3 +34,61 @@ pub fn open_test_db() -> spelunk_core::storage::Database {
...
+/// fires a foreign hook on a setup commit.
+///
+/// `/dev/null` is not a Windows path, but git skips a scope whenever its var
+/// is set, whatever the path resolves to, so this isolates on Windows too.
+///
+/// `GIT_CONFIG_GLOBAL`/`GIT_CONFIG_SYSTEM` only redirect config *files*;
+/// they don't touch `GIT_AUTHOR_*`/`GIT_COMMITTER_*`/`EMAIL`, which git
+/// consults before config and so override a test's own explicit `git config
+/// user.name`/`user.email` if the ambient process (a developer's shell, a
+/// CI runner's bot identity) happens to export them. Those are cleared too.
```
usercise: Tests should never use documenting comments `/// ` or `//! `

scripts/check-git-isolation.sh
```diff
@@ -0,0 +1,267 @@
+#!/usr/bin/env bash
+# Fails if any test code spawns `git` (`Command::new("git")`) without wiring
```
usercise: Why do we need nearly 100 lines of comments for a simple linter script?

PR #764, opened 2026-07-26
Fix two index.db data-integrity bugs: crash-half-indexed files and concurrent-run corruption

usercise:
## Summary

This closes two real, reproducible data-integrity bugs in `spelunk index`, not just added test coverage:

1. **A file interrupted mid-index stays permanently unsearchable.** `process_text_file` writes a file's new content hash (`upsert_file`) and its chunks in separate, unspanned writes. If the process is killed between the two - which a real crash-safety drill in this PR reproduces with a genuine `SIGKILL` - the file is left with a current hash and zero chunks. Because the skip check only compared hashes, every subsequent plain `spelunk index` saw "hash matches, already up to date" and silently left that file unsearchable forever, with no error and no indication anything was wrong. Only `--force` (which most users would never think to run) recovered it. Fixed by `Database::file_has_chunks` (`crates/spelunk-core/src/storage/files.rs`): the skip check now also requires the file to actually have stored chunks, so a plain re-index self-heals it on the very next run.

2. **Two concurrent `spelunk index` runs on the same project corrupt index.db.** Neither `index.db`, `memory.db`, nor `registry.db` sets `PRAGMA busy_timeout`, and SQLite's own per-connection locking does not prevent two independent processes from interleaving writes across a whole multi-transaction run. Racing two `index` processes against the same project reproducibly produced real `SQLITE_CORRUPT` ("database disk image is malformed"), not merely a benign busy/locked error. Fixed by a per-project cross-process advisory OS lock (`crates/spelunk-cli/src/cli/cmd/index/run_lock.rs`), taken as the first thing an index run does. A second run that finds the lock held now exits immediately with a clean `index already running (pid N), try again once it finishes` error instead of racing the first run's writes.

A third, related fix rides along: `Database::run_migrations` used to re-stamp `PRAGMA user_version` on every open even when nothing needed migrating, and that stamp always opens a write transaction - so a purely read-only command like `spelunk search` could fail with "database is locked" while an unrelated `spelunk index` was running. Opening an already-current database is now read-only.

## Why this matters now

Both of the two bugs above are real data-loss/corruption paths a user can hit today with an ordinary crash (OOM kill, laptop sleep, `Ctrl-C` racing a background job, a second terminal running `index` at the same time) - not edge cases requiring adversarial conditions. Bug 1 silently degrades search results with no error. Bug 2 corrupts the database outright. Both are fixed here, verified with real `SIGKILL`s against a real spawned `spelunk` binary (not simulated panics or in-process exits), and pinned by a crash-safety integration suite (`crates/spelunk-cli/tests/crash_safety.rs`, 12 tests) covering hash/chunk-write ordering, embed-phase resume, disk-full during index and memory-add, concurrent runs on the same vs. different projects, a SIGKILL'd lock holder, and a concurrent reader against an open writer transaction. Every drill asserts both `PRAGMA integrity_check` and the product's own invariants (file/chunk/embedding counts) alongside it, not just SQLite-level corruption. The concurrency test in particular self-verifies it actually exercised contention (not just two runs that happened to serialize) and was independently re-run dozens of times during development on top of its own 8-trial-per-run loop.

## Scope notes

- **Descoped this round:** a git-notes read-modify-write sequence killed mid-write is not covered by this suite. This was intentionally left out in favor of the two higher-severity bugs above (both real corruption/data-loss, vs. the git-notes case which is a narrower, lower-frequency window). Flagging it here explicitly since it isn't otherwise written down in this diff.
- **Filed separately, not fixed here:** while hardening the concurrency fix, a narrower related liveness gap was found in the lock-handoff between a parent process and a continuation child (the detached-embed / background-phase spawn path) and filed separately for a future design decision - it's a narrower edge case than the two bugs fixed here, not a regression this PR introduces.

## Test plan

- `SPELUNK_SECRET_STORE=file cargo test --workspace` - full workspace suite green (509+ tests across all crates, 0 failures).
- `SPELUNK_SECRET_STORE=file cargo clippy --workspace --all-targets -- -D warnings` - clean.
- `SPELUNK_SECRET_STORE=file cargo fmt --all -- --check` - clean.
- `SPELUNK_SECRET_STORE=file cargo test -p spelunk-cli --test crash_safety` - all 12 crash-safety drills pass.
- Independently re-ran `two_concurrent_index_runs_on_one_project_do_not_corrupt_the_db` (the highest-stakes test here, 8 trials per run, each asserting `PRAGMA integrity_check` plus exact file counts) 5 additional times end to end - all green, all with observed lock contention.

### DIFF
--- a/crates/spelunk-core/src/storage/db.rs
+++ b/crates/spelunk-core/src/storage/db.rs
@@ -32,6 +32,7 @@ impl Database {
             .with_context(|| format!("opening database at {}", path.display()))?;
 
         conn.execute_batch("PRAGMA journal_mode=WAL; PRAGMA foreign_keys=ON;")?;
+        super::apply_test_page_cap(&conn)?;
 
         let db = Self { conn };
         db.run_migrations()?;
@@ -47,12 +48,29 @@ impl Database {
     /// blindly re-running all steps would drive the guarded 008–010 ALTERs
     /// through their `duplicate column name` branch needlessly. Each step is
     /// idempotent, so a conservative (one-low) inference stays safe.
+    ///
+    /// Returns immediately, before any write, when the on-disk header
+    /// already reads `CURRENT_SCHEMA_VERSION`: `PRAGMA user_version = N`
+    /// opens a write transaction even to re-set an unchanged value, so an
+    /// unconditional stamp on every open would make every `Database::open`,
+    /// including a read-only command, contend for the write lock against a
+    /// concurrent writer instead of only doing so for a genuine migration.
     fn run_migrations(&self) -> Result<()> {
-        let mut version: i32 = self
+        // Read the RAW on-disk value before any inference below adjusts the
+        // working `version`: the skip-write gate further down must key on
+        // what is actually stamped in the file header, not on an in-memory
+        // value that was only just inferred and has not been persisted yet
+        // (a legacy pre-user_version DB reads 0 here but may infer straight
+        // to `CURRENT_SCHEMA_VERSION` - that inference still needs writing).
+        let raw_version: i32 = self
             .conn
             .query_row("PRAGMA user_version", [], |r| r.get(0))
             .context("reading user_version")?;
+        if raw_version == CURRENT_SCHEMA_VERSION {
+            return Ok(());
+        }
 
+        let mut version = raw_version;
         if version == 0 && !self.is_fresh_db()? {
             version = self.infer_legacy_version()?;
         }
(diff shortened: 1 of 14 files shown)


crates/spelunk-core/src/storage/db.rs
```diff
@@ -630,6 +654,40 @@ mod tests {
         );
     }
 
+    /// `run_migrations` used to stamp `PRAGMA user_version` unconditionally on
+    /// every open, even when nothing needed migrating - and setting that
+    /// pragma always opens a write transaction, so a concurrent reader
+    /// (`spelunk search` while `spelunk index` runs) could fail with
+    /// "database is locked" on this alone, never touching a genuine
```
usercise: Tests should never have documenting comments (`/// ` or `//! `).

usercise: Fixed: converted to a plain // comment. Checked the other 3 files this story added/touched (files.rs, crash_safety.rs, run_lock.rs) for the same pattern, all clean. Note: db.rs has many more pre-existing /// comments on other tests, predating this story, left untouched as out of scope here.

PR #780, opened 2026-07-27
index: confirm the continuation child wins the run lock before claiming the handoff succeeded

usercise:
## Summary

Stacked on #764 (`task/story-295-20260726-0940`): closes the narrower follow-on gap that PR's own description explicitly flagged as "filed separately, not fixed here" (spelunk-oss^298).

`index()` releases the per-project run lock right before spawning a continuation child (`--_embed-phases` for `--detach-embed`, `--_background-phases` for the >100-file spawn), so the child can re-acquire it itself rather than inherit a held lock. That release-then-spawn ordering closes the *corruption* race - the child never interleaves writes with the parent - but there is a narrower timing gap in the handoff itself: between the parent's release and the child's own acquire, an unrelated third `spelunk index` on the same project can win the lock first.

When that happens, the losing child bails cleanly (no corruption - already covered by PR #764's `losing_child_continuation_mode_fails_clean_without_touching_the_db` test). But the parent had already committed to telling the user its work was proceeding in the background ("Embedding N chunk(s) in the background…"), before it had any way to know whether that was true. The work the child was supposed to do is then silently dropped: not retried, and not picked up by the third process either (traced through `chunk_ids_and_texts.is_empty()`'s early return in phase 2, which only sees files *that* process itself just parsed).

## Fix

No new IPC, no FD inheritance, no DB-wide catch-up scan (all considered and rejected as overbuilding a low-probability, narrow race). Instead: reuse the lock file's existing holder-pid bookkeeping (`run_lock.rs` already writes the acquiring pid on every successful acquire, for the "already running (pid N)" error message).

After spawning a continuation child, the parent now polls (`run_lock::wait_for_holder_pid`, bounded at 2s / 20ms interval) for *that specific spawned pid* to appear as the lock's recorded holder before printing the confident "in the background" message. If it never does within the window - because a third process won the race, or the child failed to start for any other reason - the parent instead tells the user the handoff could not be confirmed and to re-run `spelunk index`, rather than asserting success it can't back up. Exit code stays 0 (consistent with the existing embed-skipped-notice precedent: a narrow miss on a background hint is not the same as the index run itself failing).

On the (overwhelming) common path, the child wins the lock within the first poll or two and the existing messaging is unchanged - verified against the pre-existing `index_background_log.rs` suite, which exercises the real happy-path spawn end to end.

## Test plan

- New unit tests for `run_lock::wait_for_holder_pid` (`crates/spelunk-cli/src/cli/cmd/index/run_lock.rs`): confirms immediately when content already matches, detects a holder that appears mid-poll (not just one present at the first check), and times out (waiting out the full budget) when the recorded pid never matches.
- New integration test `parent_reports_honest_message_when_a_third_process_wins_the_lock_handoff` (`crates/spelunk-cli/tests/crash_safety.rs`): deterministically reproduces the exact race (same pause-and-hold technique as the existing `losing_child_continuation_mode_fails_clean_without_touching_the_db` drill) by pausing the parent right after its `drop(run_lock.take())` via a new `crash_test_hook::pause_at("after_run_lock_drop", …)` point, letting a real separate `spelunk index` process win and hold the lock, then asserting on what the parent actually printed to the user (not the confident message) and that the DB stays integrity-clean.
- `SPELUNK_SECRET_STORE=file cargo test -p spelunk-cli --no-default-features` - full crate suite green (514+ tests, 0 failures), including the full `crash_safety` and `index_background_log` suites.
- `cargo fmt --all -- --check` and `cargo clippy --workspace --all-targets -- -D warnings` clean (enforced by this repo's pre-commit hook).

## Scope note

Base branch is `task/story-295-20260726-0940` (PR #764), not `main` - this fix depends on the `run_lock.rs` module #764 introduces. Merge order: #764 first, then this.

### DIFF
--- a/crates/spelunk-cli/src/cli/cmd/index/mod.rs
+++ b/crates/spelunk-cli/src/cli/cmd/index/mod.rs
@@ -228,26 +228,53 @@ pub async fn index(args: IndexArgs, cfg: Config) -> Result<()> {
     if args.detach_embed && tier.is_server() && detach_embed_eligible(&tier) {
         let embed_log = background_log_path(&db_path);
         // Release before spawning: the child re-acquires this same lock on
-        // entry, and dropping first (rather than after `spawn()` returns) is
-        // what makes that a race-free handoff instead of a timing-dependent
-        // one - the child starts running concurrently with us the instant
-        // `spawn()` is called, so releasing any later can lose the race.
+        // entry, and dropping first (rather than after `spawn()` returns)
+        // closes the corruption race - the child never interleaves writes
+        // with us. It does not by itself guarantee the *child* is what wins
+        // the reacquire: an unrelated third `spelunk index` on this project
+        // can land in the gap and take the lock first, in which case the
+        // child bails clean (see run_lock.rs) but the work it was meant to
+        // do would go silently missing. `wait_for_holder_pid` below closes
+        // that second gap by confirming the spawned pid, specifically,
+        // becomes the recorded holder before telling the user work is
+        // running in the background.
         drop(run_lock.take());
-        if let EmbedSpawn::Detached(log_in_use) =
-            spawn_embed_subprocess(&args, embed_log.as_deref())?
+        crash_test_hook::pause_at("after_run_lock_drop", "embed");
+        if let EmbedSpawn::Detached {
+            log_in_use,
+            child_pid,
+        } = spawn_embed_subprocess(&args, embed_log.as_deref())?
         {
             let stats = db.stats()?;
             let pending = stats.chunk_count - stats.embedding_count;
-            println!(
-                "Index: {} files, {} chunks. Embedding {} chunk(s) in the background\u{2026}",
-                stats.file_count, stats.chunk_count, pending,
-            );
-            if !embed_ready {
-                println!("The embedder is still loading; the background worker waits for it.");
-            }
-            println!("Run `spelunk status` to check progress.");
-            if let Some(p) = log_in_use {
-                println!("  Log: {}", p.display());
+            if run_lock::wait_for_holder_pid(
+                &spelunk_dir,
+                child_pid,
+                HANDOFF_CONFIRM_TIMEOUT,
+                HANDOFF_POLL_INTERVAL,
+            ) {
+                println!(
+                    "Index: {} files, {} chunks. Embedding {} chunk(s) in the background\u{2026}",
+                    stats.file_count, stats.chunk_count, pending,
+                );
+                if !embed_ready {
+                    println!("The embedder is still loading; the background worker waits for it.");
+                }
+                println!("Run `spelunk status` to check progress.");
+                if let Some(p) = log_in_use {
+                    println!("  Log: {}", p.display());
+                }
+            } else {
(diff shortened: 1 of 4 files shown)


crates/spelunk-cli/src/cli/cmd/index/mod.rs
```diff
@@ -228,26 +228,53 @@ pub async fn index(args: IndexArgs, cfg: Config) -> Result<()> {
...
+        // can land in the gap and take the lock first, in which case the
+        // child bails clean (see run_lock.rs) but the work it was meant to
+        // do would go silently missing. `wait_for_holder_pid` below closes
+        // that second gap by confirming the spawned pid, specifically,
+        // becomes the recorded holder before telling the user work is
+        // running in the background.
         drop(run_lock.take());
-        if let EmbedSpawn::Detached(log_in_use) =
-            spawn_embed_subprocess(&args, embed_log.as_deref())?
+        crash_test_hook::pause_at("after_run_lock_drop", "embed");
```
usercise, line 236: This is in production code, we can't call test hooks here!

crates/spelunk-cli/src/cli/cmd/index/mod.rs
```diff
@@ -228,26 +228,53 @@ pub async fn index(args: IndexArgs, cfg: Config) -> Result<()> {
...
+        // entry, and dropping first (rather than after `spawn()` returns)
+        // closes the corruption race - the child never interleaves writes
+        // with us. It does not by itself guarantee the *child* is what wins
+        // the reacquire: an unrelated third `spelunk index` on this project
+        // can land in the gap and take the lock first, in which case the
+        // child bails clean (see run_lock.rs) but the work it was meant to
+        // do would go silently missing. `wait_for_holder_pid` below closes
+        // that second gap by confirming the spawned pid, specifically,
+        // becomes the recorded holder before telling the user work is
+        // running in the background.
```
usercise: This file has so mainly lines of useless comment slop that it's completely impossible to follow the code logic. Please clean up.

crates/spelunk-cli/src/cli/cmd/index/mod.rs
usercise, line 1: This file is too long, please set up another PR to refactor it.

crates/spelunk-cli/tests/crash_safety.rs
```diff
@@ -1058,3 +1055,83 @@ fn losing_child_continuation_mode_fails_clean_without_touching_the_db() {
     assert_integrity_ok(&db_path);
     kill_and_reap(paused);
 }
+
+#[test]
+fn parent_reports_the_handoff_honestly_when_a_third_process_wins_the_lock_race() {
+    // The gap the previous test's drill doesn't cover: it drives the losing
+    // child directly, never the real parent-releases/parent-spawns handoff,
+    // so it cannot see what the *parent* tells the user. Before this fix, the
```
usercise: Stop talking about what the code used to be in comment, that is completely useless to a new engineer that starts a week from now, it doesn't help them understand the code. This is just extra context they have to carry for now reason.

PR #786, opened 2026-07-27
fix: keep the local server responsive while it is embedding

usercise:
During a `spelunk index` run the local `spelunk-server` became unreachable for
the whole embed phase. `/v1/health` took seconds instead of well under a
millisecond, `spelunk server status` reported a perfectly healthy server as
`(unreachable)`, and endpoints with nothing to do with embedding stopped
answering too. Everything recovered the instant the index completed.

## Cause

`/v1/health` read the embedder's per-chunk token cap through
`EmbeddingBackend::token_cap()`, and `NativeEmbedder::token_cap()` locked the
same `Arc<Mutex<EmbedderInner>>` that `embed_with_cancel` holds for a request's
entire batch (up to 32 sequential forward passes). Because that lock is
synchronous inside an `async fn`, it blocked a tokio worker rather than
yielding it, so each concurrent probe parked another worker until tasks for
unrelated endpoints could no longer be polled. That is the escalation from
"health is slow" to "the whole server is unreachable".

## Fix

The cap is derived once at model load and never mutated, so it is now a plain
field on `NativeEmbedder` and the accessor is a field read. `embed_with_cancel`
resolves the effective cap before entering its `spawn_blocking` closure.
Structurally there is no longer any path from a liveness probe to the
forward-pass mutex.

The embedder's own behaviour is untouched: batches are still serialized behind
that mutex, the thread budget is unchanged, and admission control is unchanged.

## Compatibility

The `/v1/health` wire format is unchanged and `limits.embedder_token_cap`
reports the same value as before, so existing clients need no change. The
health handler itself is byte-for-byte identical.

## Tests

Three deliberately separate groups:

- `spelunk-embed` pins the accessor against a genuinely held forward-pass
  mutex, using a fixture embedder that needs no model on disk. The read runs on
  its own thread behind a bounded `recv_timeout`, so a re-coupled accessor
  fails the bound instead of hanging the suite. These two tests are the
  regression guard, and the fixture builds the struct by literal so moving the
  field back fails at compile time.
- `spelunk-server` is the deterministic gate: a mock embedder parks inside
  `spawn_blocking` holding a forward-pass mutex on a condvar gate, so "an embed
  is in flight" is an asserted state rather than a timing race. No model, GPU,
  repo, network, or sleeps. Covers a single probe, eight concurrent probes
  (worker-thread exhaustion), the full response body, an embedder-free
  endpoint, `memory search` queueing behind the batch, recovery after
  completion, and a cancelled embed. One test proves the 250 ms bound actually
  catches the coupling it guards against.
- An `#[ignore]`d real-hardware run samples `/v1/health` every 500 ms across a
  real embed load, asserting every sample stays under the bound with a stable
  server identity. Explicitly not a CI gate.

The module comment states plainly that the server-side tests run against a mock
and therefore stay green if the real accessor is re-coupled, and names the two
tests that do guard it.

## Docs

The "Embedding CPU thread budget" section of `docs/server-setup.md` justified
the knob by claiming a CPU-only host's embed batch could make `/v1/health` go
unresponsive. The symptom was real but the cause was not CPU fan-out, so an
operator following that rationale would have tuned a knob that could never have
helped. The section now describes what the budget actually bounds and says
outright that it is not what keeps liveness probes answering.

### DIFF
--- a/crates/spelunk-server/src/handlers.rs
+++ b/crates/spelunk-server/src/handlers.rs
@@ -4842,8 +4842,20 @@ mod tests {
     // race. `cap_location` is the mock's one degree of freedom, and
     // `harness_detects_a_cap_read_behind_the_forward_pass_mutex` uses it to
     // prove these bounds actually catch the coupling they guard against.
-    // The real `NativeEmbedder` accessor is pinned in `spelunk-embed`, which a
-    // mock cannot cover.
+    //
+    // Know what this module does NOT do before you rely on it. Every test here
+    // runs against `ParkingEmbedder`, so re-coupling the real
+    // `NativeEmbedder::token_cap()` to its forward-pass mutex leaves all of
+    // them green: verified by mutation, all pass with the accessor put back
+    // behind the lock. That is structural and not fixable here, because these
+    // tests cannot construct a `NativeEmbedder` without a model on disk. What
+    // this module gates is the server-side property (health, and endpoints
+    // that need no embedder, stay prompt given a backend whose cap read is
+    // lock-free) plus the sensitivity of the bound itself. The regression
+    // guard for the accessor is `spelunk_embed::embedder_native`'s
+    // `token_cap_returns_while_the_forward_pass_lock_is_held` and
+    // `token_cap_is_independent_of_embedder_busyness`. If you change the cap's
+    // storage, those are the tests that must fail.
     mod liveness_under_embed {
         use std::sync::{Arc, Condvar, Mutex};
         use std::time::{Duration, Instant};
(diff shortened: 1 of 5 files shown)


crates/spelunk-server/src/handlers.rs
```diff
@@ -4825,4 +4825,484 @@ mod tests {
         // this not-cancelled can't turn this test into a silent hang.
         release.notify_one();
     }
+
+    // ── Liveness while an embed is in flight ─────────────────────────────────
```
usercise, line 4829: This comment is bonkers too long, and that is inside a file that is already long.

PR #790, opened 2026-07-28
memory push/sync: embed the push set locally before shipping it

usercise:
`spelunk memory push` and `spelunk sync` now embed every entry in the push set
that lacks a usable local vector, through the local loopback embedder, and
commit each vector to `memory.db` before the batch is built. Previously a push
left the local store exactly as it found it: it reported `created`, and the rows
it had just pushed stayed invisible to semantic `spelunk memory search` on the
user's own machine, with nothing saying so and no hint that
`spelunk memory reindex` was the cure.

## What travels is unchanged

`kind`, `title`, and `body` were always serialised on every push, and the
optional `vector` / `vector_model` / `vector_precision` fields are additive. A
local embedding changes zero bytes of entry content on the wire. What changes is
that the local store is left correct, and that a destination advertising
`accepts_pushed_vectors` can store the entry as-is instead of re-embedding it.
The existing `maybe_attach_vector` gate is untouched.

## The guarantee

The local embed resolves through the inference tier (`get_inference_tier` plus
`effective_config`), exactly as `memory reindex` does. Outside `cloud_first`
that branch probes loopback and nothing else, and `resolve_inference_url` falls
back to `server_url` only under `cloud_first`, which is precisely the case the
repair skips. A configured team `server_url` is therefore structurally
unreachable from the embed path, so the fix cannot recreate the server-side
re-embedding it removes.

The document string and the document-side call are byte-identical to
`memory reindex`'s (`title: {title} | text: {body}`, via `embed_text`, never a
query-side prefix), so a vector minted during a push is interchangeable with a
reindexed one.

## Behaviour at the edges

- No local embedder reachable: the push runs to completion text-only with the
  exit code it always had, emits one warning naming how many entries went out
  unembedded and the remedy, and reports the count in the summary line.
- A single row's embed failure: that row ships text-only, the rest still get
  vectors, and the push does not abort.
- Skipped entirely in `cloud_first` with a `server_url` set, the same condition
  `memory reindex` declines under. The two are now pinned against each other
  across six mode / `server_url` shapes so neither can drift alone.
- Rows already synced, and archived rows, are outside the push set and are not
  embedded.
- Vectors commit per row, so an interrupted push keeps every vector it minted
  and a re-run embeds only the remainder.

## Docs

`docs/memory.md` and `docs/commands.md` both stated that a note's semantic
vector is minted only at `memory add` time. That is no longer true, and both are
corrected. A new "Repair during push and sync" section covers the behaviour and
its limits.

### DIFF
--- a/crates/spelunk-cli/src/cli/cmd/memory/sync/push/embed_routing_tests.rs
+++ b/crates/spelunk-cli/src/cli/cmd/memory/sync/push/embed_routing_tests.rs
@@ -1,0 +1,407 @@
+//! Where the pre-batch local embed is allowed to go, and what happens when it
+//! cannot run at all.
+//!
+//! The load-bearing invariant here: the embed must reach the loopback embedder
+//! and never the configured team `server_url`. Routing it there would re-create
+//! the exact server-side re-embedding the repair exists to remove.
+
+use super::super::test_support::{fresh_store, spawn_loopback_embedder};
+use super::*;
+use crate::config::{Config, SyncMode};
+
+use wiremock::matchers::{method, path};
+use wiremock::{Mock, MockServer, ResponseTemplate};
+
+fn add_note(store: &MemoryStore, title: &str) -> String {
+    store
+        .add_note("decision", title, "body", &[], &[], None, None)
+        .unwrap();
+    let rows = store.rows_for_sync(true).unwrap();
+    rows.iter()
+        .find(|r| r.title == title)
+        .expect("note added")
+        .uuid
+        .clone()
+}
+
+async fn mount_batch_created(server: &MockServer, uuid: &str) {
+    Mock::given(method("POST"))
+        .and(path("/v1/projects/proj/memory/batch"))
+        .respond_with(ResponseTemplate::new(207).set_body_json(serde_json::json!({
+            "created": 1, "skipped": 0, "failed": 0,
+            "results": [{"status": "created", "external_id": uuid, "id": "cloud-1"}]
+        })))
+        .mount(server)
+        .await;
+}
+
+fn paths(reqs: &[wiremock::Request]) -> Vec<String> {
+    reqs.iter().map(|r| r.url.path().to_string()).collect()
+}
+
+// ── the never-route-to-server_url invariant ─────────────────────────────────
+
+#[tokio::test]
+#[serial_test::serial]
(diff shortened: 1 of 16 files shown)


crates/spelunk-cli/src/cli/cmd/memory/sync/push/embed_routing_tests.rs
```diff
@@ -0,0 +1,518 @@
+//! Where the pre-batch local embed is allowed to go, and what happens when it
+//! cannot run at all.
+//!
+//! The load-bearing invariant here: the embed must reach the loopback embedder
+//! and never the configured team `server_url`. Routing it there would re-create
+//! the exact server-side re-embedding the repair exists to remove.
```
usercise: Never use docs-comment for tests. That rule is in CLAUDE.md.

crates/spelunk-cli/src/cli/cmd/memory/sync/push/embed_scope_tests.rs
```diff
@@ -0,0 +1,257 @@
+//! What the pre-batch local-embedding repair deliberately does NOT touch: rows
```
usercise: docs-comment again.

crates/spelunk-cli/src/cli/cmd/memory/sync/push/embed_tests.rs
```diff
@@ -0,0 +1,396 @@
+//! Pre-batch local-embedding repair tests for [`super::push_local`]: a pushed
```
usercise: again

crates/spelunk-cli/src/cli/cmd/memory/sync/test_support.rs
```diff
@@ -52,6 +52,103 @@ pub(super) async fn spawn_spelunk_server() -> std::net::SocketAddr {
...
+            }
+            match self.prev_no_server.take() {
+                Some(v) => std::env::set_var("SPELUNK_NO_SERVER", v),
+                None => std::env::remove_var("SPELUNK_NO_SERVER"),
+            }
+        }
+    }
+}
+
+/// The fp32 vector [`spawn_loopback_embedder`]'s `/index/embed` route returns.
```
usercise: No docs-comment for tests.

PR #797, opened 2026-07-28
Configurable LLM endpoint, model and credential, wired through to the spawned server

usercise:
Make the LLM endpoint a first-class client-side setting, and give its credential a place to live that the detached daemon can never reach into.

## What this adds

- `llm_url` and `llm_model` in the **personal** config (`~/.config/spelunk/config.toml`), overridable by `SPELUNK_LLM_URL` / `SPELUNK_LLM_MODEL`, and per launch by `spelunk server start --llm-url/--llm-model`.
- `spelunk auth set-key --llm` stores the endpoint's credential in the OS secret store, read from stdin or an interactive prompt.
- `spelunk-server` gains `--llm-key` and `--llm-key-file` alongside `SPELUNK_LLM_KEY`, and sends a resolved credential as a bearer token upstream.

Previously only an exported environment variable reached the auto-spawned server, because the CLI passed it nothing but host, port and database path. A `config.toml` setting was silently dropped, and there was no way to configure a credential at all, so a keyed endpoint could not be used.

## Security properties

These are the reasons the change is shaped the way it is, and each has a test.

- **The credential never travels in an argument.** No input causes `--llm-key` or `--llm-key-file` to be emitted into the spawned daemon's argv, because argv is world-readable through the process table. The endpoint URL and model do travel as arguments: neither is secret, and `ps` showing which endpoint a daemon serves is a diagnostic feature.
- **The spawned daemon never opens the OS secret store.** It is detached and long-lived, and a keychain read from a background process with no user session raises an authorization prompt nobody can answer. The CLI resolves the credential in the user's own session and passes it out of band, in the child's environment. Enforced by a source-level scan over the whole server crate, so a future reach fails CI rather than shipping.
- **Resolving the credential costs nothing on other commands.** It is not a config field and is not read when configuration loads; only the daemon-spawn path reads it. An ordinary `spelunk search` does not authorize against the keychain for it.
- **A credential is not sent over plaintext to another host.** When a credential resolves and the endpoint is a plaintext `http://` URL to a non-loopback host, the server refuses to start, naming the URL, before it opens the database. The check is scoped to a credential being present, so an existing keyless LAN endpoint (LM Studio or Ollama) keeps working unchanged, and a keyless endpoint is still sent no `Authorization` header at all.
- **`llm_url` is personal-config only.** A value in a checked-in `.spelunk/config.toml` is ignored, so cloning a repository cannot point a developer's local daemon, or their credential, at someone else's machine.

## Determinism of the daemon's view

All three LLM variables are set on the child rather than inherited from the parent shell, so what the CLI resolved is exactly what the daemon sees. The server reads `SPELUNK_LLM_URL` and `SPELUNK_LLM_MODEL` itself, so a value the CLI deliberately resolved away would otherwise still reach it behind its back: an exported `SPELUNK_LLM_URL=""` would arrive as a configured-but-empty endpoint rather than as no endpoint.

## Also fixed

`spelunk server start` no longer blames the firewall for a daemon that refused its own startup configuration. It now watches the child, reports an exited process immediately and points at the log; the firewall wording is kept only for the case it actually describes, a daemon still running and still not answering.

## Compatibility

A daemon that is already running keeps the configuration it started with; restart it to pick up a change. Nothing restarts it automatically. Existing keyless setups are unaffected, and an unconfigured spawn produces a byte-identical argument list to before. No new dependencies.


CHANGELOG.md
```diff
@@ -11,6 +11,56 @@ spelunk uses [Semantic Versioning](https://semver.org/).
...
+  --llm-url/--llm-model`, and store the endpoint's credential with `spelunk auth
+  set-key --llm` (read from stdin or a prompt, kept in the OS secret store,
+  overridable with `SPELUNK_LLM_KEY`). Previously only an exported environment
+  variable reached the auto-spawned `spelunk-server`, because the CLI passed it
+  nothing but host, port, and database path: a `config.toml` setting was silently
+  dropped, and there was no way to configure a credential at all, so a keyed
+  endpoint could not be used. Both keys are read from the **personal** config
+  only; a value in a checked-in `.spelunk/config.toml` is ignored, so cloning a
+  repository cannot point your local daemon at someone else's machine.
+
```
```suggestion
- **A configurable LLM endpoint and a  place to keep its credential.** You can now set `llm_url` and `llm_model` in
  `~/.config/spelunk/config.toml`, override either with `SPELUNK_LLM_URL` /
  `SPELUNK_LLM_MODEL` or per launch with `spelunk server start
  --llm-url/--llm-model`, and store the endpoint's credential with `spelunk auth
  set-key --llm` (read from stdin or a prompt, kept in the OS secret store,
  overridable with `SPELUNK_LLM_KEY`).

```

CHANGELOG.md
```diff
@@ -11,6 +11,56 @@ spelunk uses [Semantic Versioning](https://semver.org/).
...
+  The credential is not a config field and is not read when configuration loads.
+  Only the code path that spawns the daemon resolves it, so no ordinary command
+  authorizes against your keychain for it. The daemon itself never opens the
+  secret store, which as a detached process with no user session it could not do
+  without raising a prompt nobody can answer: the CLI resolves the value and
+  hands it over out of band, in the child's environment. The endpoint URL and
+  model travel as ordinary arguments, since neither is secret and seeing which
+  endpoint a daemon serves in `ps` is useful, but **no input causes the
+  credential to appear in an argument**, and none of this prints it.
```
```suggestion
```

CHANGELOG.md
```diff
@@ -11,6 +11,56 @@ spelunk uses [Semantic Versioning](https://semver.org/).
...
+  credential is still sent no `Authorization` header at all.
+
+  What the CLI resolves is exactly what the daemon it starts sees. All three LLM
+  variables are set on the child rather than inherited from your shell, so an
+  exported `SPELUNK_LLM_URL=""` genuinely switches the endpoint off for that
+  shell instead of reaching the daemon as a configured-but-empty one.
+
+  A daemon that is already running keeps the configuration it started with;
+  restart it with `spelunk server stop && spelunk server start` to pick up a
+  change. Nothing restarts it for you.
```
```suggestion
```

PR #798, opened 2026-07-28
Version-skew support policy, a real-binary skew smoke test, and independent health-field degradation

usercise:
## Why

`docs/stability.md` says what a surface promises *within* one version. Nothing said what
happens when the two ends of a connection are *different* versions, which is the normal case:
the CLI talks to an auto-discovered loopback server, a team `spelunk-server` upgraded on
someone else's schedule, and a continuously deployed hosted API, and all three drift
independently.

Wire tolerance was previously tolerance by accident. This makes it a stated policy, gives it
tests against real artifacts, and fixes a live forward-compatibility defect those tests found.

## What this adds

**A support policy: `docs/version-skew.md`.** The supported window per peer (a team
`spelunk-server` one minor version either side, a loopback server at the same version, the
hosted API at any version because it evolves additively within `/v1/`), what holds outside
that window, and, just as important, what is *not* promised. Outside the window the CLI keeps
working on a best-effort basis rather than refusing to run. That is deliberate: a hard version
gate turns every "upgrade the CLI before the server" ordering into an outage for the person
doing the upgrade. The document states plainly that the CLI does not compare its version
against the peer's at all today, so nothing in it is enforced by a version check.

**A binary skew smoke test: `scripts/skew-smoke.sh` and a `version-skew` workflow.** It puts
the current build and the previous released build on a socket together, in both directions,
and drives the whole memory flow: add, list, search, push, re-push, sync, and pull into a
fresh checkout. Every other contract test in this repo talks to a mock written to the shape we
believe a peer has, so this is the only one that can contradict that belief rather than confirm
it. The workflow verifies each downloaded release asset against the digest the release
publishes before executing it. Two guards stop the job going green while proving nothing: it
refuses to run two identical versions against each other, and it fails rather than skipping if
the server-side embedder never became ready, because search is the only step that drives the
query-embedding path across the skew boundary.

The script reduces the child environment to an allowlist before any binary starts. Everything
not named is unset, so a path override added to the tree later is excluded by default rather
than by memory. That matters because the binaries under test are old releases that pre-date the
file secret-store default: an inherited `SPELUNK_CONFIG_DIR` would have pointed them at the
developer's real `secrets.toml`, to read and to write.

**Peer-to-CLI fixtures recorded from real released binaries**, not written to our assumptions.
The v0.8.0 and v0.9.0 bodies genuinely omit `embedder`, `embedding_dim` and `limits`, so the
absent-optional path is now exercised by a peer that really behaved that way. The document
records the provenance of each fixture and says which are recordings and which are not.

## The behaviour change

**One unreadable field in a health response no longer costs every capability the server
advertised.** If any single value in `GET /v1/health` was in a shape this build could not
parse, the whole response was discarded and the peer treated as a legacy plain-text server:
semantic search, index embed and harvest all reported unavailable, advertised limits dropped,
embedding dimension lost, and nothing logged to say why. It took very little to trigger. A
newer server adding one value to an existing enum, which the stability contract expressly
permits as an additive change, was enough; so was a field holding an explicit `null`, which
this server family emits routinely.

Every member now degrades on its own, including inside the nested `limits` object, so a server
advertising `max_batch_chunks: 16` next to one unreadable sibling keeps the 16 instead of
leaving the client planning around its own maximum of 256. The guarantee is enforced by a test
that mutates every member of a recorded response in turn, so a member added later without a
lenient read fails the same check.

Degrading is also no longer silent. The CLI warns, naming the field, the shape it expected and
what the fallback costs. Those lines deliberately never render the field's value: `/v1/health`
needs no authentication and its body is whatever `server_url` resolves to, so its contents are
peer-controlled and unbounded, and a diagnostic line is no place to reproduce them. The
warnings are `warn`-level and the CLI logs at `error` by default, so `RUST_LOG=warn spelunk
status` is what surfaces them.

## Scope, stated rather than implied

- No version gate ships. Whether the CLI should ever refuse a pairing outright, and whether a
  mismatched loopback server should warn or refuse, are open product decisions; the document is
  a policy statement rather than a description of enforcement.
- Only the response direction is validated against a schema. The recorded previous-release
  OpenAPI spec is committed and ready for request-side validation, and nothing consumes it yet.
- The hosted API is not represented in this repository, so the cross-peer divergence table in
  the document is a transcribed note that will go stale silently, and is labelled as such at the
  table itself.
- A query parameter the CLI sends that the server accepts and then ignores is pinned as a test
  rather than fixed, with the consequence spelled out where the pin lives.
- `crates/spelunk-cli/tests/fixtures/skew/openapi-v0.9.4.json` is a verbatim recording of a
  released binary's own output and is exempt from the house punctuation and reference
  conventions. Editing it to conform would destroy the only property it has, and the document
  explains this in place.

## Verification

Run without exporting `SPELUNK_CONFIG_DIR`, which currently trips an unrelated defect on `main`.

- `cargo fmt --all --check`, `cargo clippy --all-targets --features rich-formats -D warnings`,
  `scripts/check-git-isolation.sh`, `bash -n scripts/skew-smoke.sh`: all clean.
- Full workspace `cargo test` including doctests: 1941 passed, 0 failed.
- `cargo audit`: no vulnerabilities, one pre-existing allowed warning. No dependencies added.
- The warning paths were attacked with hostile peer bodies beyond those the tests pin: non-UTF8
  bytes, an object with a 100 kB key, five thousand unknown credential-bearing members, a
  credential nested a hundred levels inside `limits`, and a `limits` object replaced by a 100 kB
  string. Every one stays under 8 kB of log and renders no peer credential.
- The environment allowlist was attacked with more than forty hostile variables covering config,
  state, registry, model, secret store, `CREDENTIALS_DIRECTORY`, the `GIT_*` family, `LD_PRELOAD`,
  the `DYLD_*` family, the Hugging Face variables, the proxy variables and the TLS trust
  variables. None reaches the child, and the backstop assertion exits 1 on an unowned export.


CHANGELOG.md
```diff
@@ -50,6 +50,33 @@ spelunk uses [Semantic Versioning](https://semver.org/).
   for the full fetch-and-transfer procedure. Unset by default, so the online
   path is unchanged for the common case.
 
+- **A version-skew support policy, [docs/version-skew.md](docs/version-skew.md),
```
usercise, line 108: A changelog entry should be a few words, a paragraph at most. It's not an invite to write a life story.

CHANGELOG.md
```diff
@@ -105,6 +105,33 @@ spelunk uses [Semantic Versioning](https://semver.org/).
...
+  request direction is not validated against any peer schema yet, and the
+  hosted API is not represented in this repository at all, so the divergence
+  table in the document is a transcribed note that will go stale silently
+  rather than a check. A new `version-skew` workflow puts the current build and
+  the previous release on a socket together in both directions and drives the
+  full memory flow (add, list, search, push, sync, and pull into a fresh
+  checkout), verifying each downloaded release asset against its published
+  digest before running it. The job refuses to pass if the two binaries turn out
+  to be the same version, or if the embedder never became ready so the search
+  step never ran: both would go green while proving nothing.
```
```suggestion
- **A version-skew support policy, [docs/version-skew.md](docs/version-skew.md),
  and a CI job that runs two real binaries against each other.** A new `version-skew` 
  workflow puts the current build and the previous release on a socket together in 
  both directions and drives the full memory flow (add, list, search, push, sync, and 
  pull into a fresh checkout), verifying each downloaded release asset against its 
  published digest before running it.
```

CHANGELOG.md
```diff
@@ -132,6 +159,33 @@ spelunk uses [Semantic Versioning](https://semver.org/).
...
+  maximum of 256. The guarantee is now that no single field can take the whole
+  body down, and it is enforced by a test that mutates every member of a
+  recorded server response in turn, so a field added later without a lenient
+  read fails the same check. Degrading is also no longer silent: the CLI warns,
+  naming the field, the shape it expected, and what the fallback costs. Those
+  warnings are `warn`-level, so run `RUST_LOG=warn spelunk status` to see them.
+  They deliberately never print the field's value: `/v1/health` needs no
+  authentication and its body is whatever `server_url` resolves to, so its
+  contents are peer-controlled, and a diagnostic line is no place to reproduce
+  them.
```
```suggestion
- **One unreadable field in a server's health response no longer costs you
  every capability that server advertised.** The CLI reads `GET /v1/health` to
  learn what a server can do. The guarantee is now that no single field can take
  the whole body down, and it is enforced by a test that mutates every member 
  of a recorded server response in turn.
```

PR #799, opened 2026-07-29
Route LLM inference separately from embedding, so index summaries run

usercise:
> **Stacked PR.** This targets `task/engineer-story301-llmconfig-20260727-2212`
> (open as #797, not yet merged), **not `main`**. It reads the `llm_url` config
> that branch adds. Retarget to `main` once #797 merges; review the diff of this
> PR alone, not the combined range.

`spelunk index` skipped every chunk summary. Two gates did it: a project with no
`server_url` was skipped outright, and a project *with* one failed while
reporting an internal type name. Between them the summary pass was off for
everyone, and the more common loopback-only case never even reached the line the
error came from. Both gates are gone.

## What changed

LLM inference now resolves on its own rule, separate from embedding:

1. Offline mode: nothing runs, and nothing is probed.
2. Your local server serves an LLM: it is used.
3. `llm_url` is set but your local server does not serve an LLM: stop and ask
   for a restart (see below).
4. A configured `server_url` serves an LLM: it is used.
5. Otherwise: no LLM is available, and the message says so with both ways to get
   one.

Availability comes from the `llm.complete` capability the reachable server
advertises, not from a config field, because a field cannot tell you whether the
running daemon ever picked the value up. `explore` and all three `memory
harvest` sources follow the same rule; each now builds one client for the LLM
and one for embedding, since the two can resolve to different servers in the
same command.

`spelunk index` skips and exits 0, because summaries are optional
(`--no-summaries` silences the notice). `explore` and `memory harvest` fail,
because neither can do its job without an LLM. One shared message renderer over
(reason x caller) so the three commands read as one product, and no message
names an internal type, adapter or field.

## The local-LLM guarantee, and where it stops

**Under the default `local_first` mode and under `offline`: if `llm_url` is set,
your code is never sent to a remote LLM.** When the running local server is not
serving that endpoint, the command stops and asks you to restart it rather than
quietly substituting `server_url`. The message for that case deliberately never
mentions `server_url`, so it cannot nudge you toward the very thing you did not
choose.

**Under `mode = "cloud_first"` this does not apply.** There the configured
`server_url` already *is* your inference target, so it is matched at rule 2 and
rule 3 never runs: LLM calls go to that server even with `llm_url` set. That is
consistent rather than an escape hatch, since `cloud_first` already routes
embedding to the same server, but it is a real boundary and it is stated as such
in the docs and the CHANGELOG rather than left implicit.

## Embedding routing is untouched

`spelunk-core` has no changes on this branch at all, so the embed resolver
cannot have moved. Its unit tests and `tests/index_embed_tier_routing.rs` pass
unmodified.

## Verification

Full workspace `cargo test` green (92 binaries, 2065 tests, exit 0); fmt, clippy
`-D warnings`, doctests, git-isolation and `--no-default-features` all green.
Routing was mutation-tested at every layer it passes through: the capability
parse, all four resolver steps, the effective-config target, the explicit-remote
constructor, and each call site individually, including all three harvest
sources separately. Every mutation is red.

Both forms of the reported symptom were reproduced end to end against mock
servers, along with the guarantee and its `cloud_first` boundary, by driving the
real binary rather than by unit test alone.


CHANGELOG.md
```diff
@@ -26,8 +43,44 @@ spelunk uses [Semantic Versioning](https://semver.org/).
...
+  exported `SPELUNK_LLM_URL=""` genuinely switches the endpoint off for that
+  shell instead of reaching the daemon as a configured-but-empty one.
+
+  A daemon that is already running keeps the configuration it started with;
+  restart it with `spelunk server stop && spelunk server start` to pick up a
+  change. Nothing restarts it for you.
+
 ### Fixed
 
+- **`spelunk index` no longer skips every chunk summary, and says something you
```
usercise, line 31: This needs to be rewritten to focus on what the user cares about and made much shorter.

PR #801, opened 2026-07-29
Serve cloud_first memory against the hosted API

usercise:
## What changed and why

`mode = "cloud_first"` could only ever speak one memory dialect, the self-hosted `spelunk-server`'s. Pointed at the hosted API, every CRUD call failed: the routes differ, and the entry shape differs beneath them, so ids failed to deserialize before any route question was reached. Remote agents need a project whose memory has no local store at all, because running a full index before each session is not feasible for them, and that is the gap this closes.

Three pieces:

**1. A memory entry's id is opaque at the backend boundary.** The trait threaded `i64` through `add`/`get`/`archive`/`supersede`, which only a rowid-keyed store can satisfy. Identity is now a `NoteId` newtype. The SQLite schema, migrations, dedupe, sync and `MemoryEdge` all stay keyed on `i64` and narrow at their own edges, so this is a boundary change rather than a rewrite. An id aimed at a store that cannot resolve it now says so instead of reporting the entry as missing.

`--format json` output is unchanged: a numeric id still serializes as a JSON number, and only a non-numeric id serializes as a string. Quoting every id unconditionally would have broken existing consumers for the local and self-hosted populations, which is every current user, in order to describe a UUID none of them have.

**2. The dialect is settled once, when the backend opens**, from the capability list `/v1/health` already advertises. A peer offering SSE memory streaming is the hosted API; every other answer, including a probe that times out or returns no capability list, selects the self-hosted dialect. That fallback is what keeps a self-hosted deployment on exactly its previous code path. The probe is unauthenticated. No new health field was needed on either peer.

**3. Two concrete backend types, not branches inside one.** The self-hosted dialect keeps its existing type untouched, which is what makes "self-hosted `cloud_first` is unchanged" true by construction rather than by care.

## Acceptance criteria and evidence

| Criterion | Evidence |
|---|---|
| Peer advertising SSE streaming selects the hosted dialect | `peer::tests::memory_stream_capability_selects_the_cloud_dialect`, `backend_selection_tests::a_health_probe_advertising_memory_stream_selects_the_cloud_dialect` |
| Self-hosted capability shape keeps today's dialect | `team_server_capabilities_select_the_team_dialect`, `a_team_server_health_probe_keeps_the_self_hosted_dialect` |
| Unreachable / legacy / capability-less health keeps today's dialect | `an_unreachable_probe_falls_back_to_the_team_dialect`, `a_legacy_plain_text_health_body_...`, `a_health_body_without_capabilities_...`, `a_non_success_health_status_...`, `an_unanswerable_health_probe_keeps_the_self_hosted_dialect` |
| The probe never sends a bearer | `the_probe_never_sends_a_bearer` (the mock answers with the cloud capability iff it saw an `Authorization` header, so a leak flips the assertion) |
| Ids round-trip in both shapes; JSON output unchanged | `note_id::tests` (8 cases, incl. `numeric_ids_serialize_as_json_numbers`) |
| An id from the wrong store gives an actionable error, not a panic | `backend::numeric_note_id` |
| add / list / search / count against the hosted API | `add_posts_to_the_memory_route_and_returns_the_server_uuid`, `list_reads_the_memory_route_without_a_query`, `search_uses_the_query_parameter_on_the_list_route`, `count_reads_the_server_computed_total`, `an_empty_project_lists_cleanly` |
| Every add carries an `external_id` | `add_always_mints_an_external_id` |
| `--mode text` keeps the shared remote message | `bm25_text_search_keeps_the_shared_remote_message` |
| get: 404 is `None`, tombstone becomes archived | `a_missing_entry_is_none_not_an_error`, `an_archived_tombstone_becomes_archived_status` |
| archive via DELETE, 404-idempotent | `archive_deletes_rather_than_posting_an_archive_subroute`, `archiving_an_already_gone_entry_is_not_an_error` |
| supersede via batch edge; unresolved is not reported as success | `supersede_posts_a_batch_edge_keyed_by_external_id`, `an_unresolved_edge_reports_no_change_rather_than_success` |
| A slug `project_id` works on every route, including get/archive | `a_slug_project_gets_an_entry`, `a_slug_project_archives_an_entry`, `a_slug_project_still_lists_and_adds` |
| Harvest dedupe filters client-side across a page boundary | `harvested_shas_pages_past_the_first_full_page`, `has_source_ref_matches_on_source_commit` |
| 409 on add is a warning, not a failure | `a_conflict_on_add_is_a_warning_not_a_failure` |
| Self-hosted dialect unregressed | all pre-existing `remote::tests` and `backend_selection_tests` pass with only the type substitution; no route, verb or payload changed for that peer |

## Gates

Run on `5415985` with `SPELUNK_SECRET_STORE=file` on every cargo command and `SPELUNK_CONFIG_DIR=$(mktemp -d)` on test runs.

- `cargo fmt --all -- --check`: pass
- `cargo clippy --all-targets --features rich-formats -- -D warnings`: pass, zero diagnostics
- `cargo test --workspace --features rich-formats`: `spelunk-core` lib and `spelunk-cli` unit suites fully green
- `cargo test --doc --workspace`: pass
- `scripts/check-git-isolation.sh`: OK
- Source hygiene on the diff: no tracker references, no em-dashes, no doc-comments in tests

`cargo-nextest` is not installed in the environment these were run in, so the suite ran under `cargo test`.

Eight test binaries have failures that reproduce identically on `main` with the same command and environment (`adr037_p2_auto_start_scope` 5, `e2e_cli` 2, `memory_push_sync_total_failure` 4, `memory_read_mode_notice` 3, `memory_reconcile` 1, `plumbing_jsonl_contract` 1, `pre_push_hook` 1, `read_memory` 3 vs 5 on main). Each was measured against `main` in the same container before and after; nothing in this branch changes their count. They appear to be environment-sensitive, so a cleaner environment may show fewer.

Two notes for reviewers on deliberate changes to existing assertions:

- `assert_no_preflight` forbade `GET /v1/health` at backend open. That guard was introduced to stop the retired slug-to-UUID resolver (`GET /v1/projects`) from coming back, whose failure took down every memory command for the self-hosted population. This branch's `/v1/health` probe cannot cause that harm, because every failure to answer it resolves to the self-hosted dialect. The assertion is narrowed to the project lookup, which is the harm it was actually protecting, and the mocks in those tests still mount no `/v1/health`, so each of them now also exercises the fallback.
- Three tests pinned an exact request list and now filter the probe out. They still assert that no project lookup happens and that the memory call is the only memory request.

## Known follow-ups, not blocking

- The hosted API has no server-side `source_commit` filter, so harvest's duplicate check pages and filters client-side: correct, but proportional to project size rather than an indexed lookup. Worth a server-side filter later.
- The peer probe's timeout is 5s. It is a liveness probe against a server the client is about to hit with a real 30s-timeout request, so ~2s would do. Worth folding into the next commit that touches this branch.
- Separately raised in review: the self-hosted server's note response carries no `valid_at` / `invalid_at` / `source_ref`, so it never reports when an entry was archived, even though the CLI's wire type already declares all three as optional and the domain type carries them. Purely additive to fix, and independent of this PR.

## Docs

CHANGELOG entry, a `docs/memory.md` section on dialect detection and the client-side harvest filtering cost, and a second amendment to ADR-005 closing out the per-entry-route constraint it flagged.


crates/spelunk-core/src/storage/remote/cloud_api/wire.rs
```diff
@@ -0,0 +1,138 @@
+//! Wire types for the cloud API's memory routes.
+//!
+//! Deliberately separate from the team server's types in
+//! [`wire_types`](super::super::wire_types): the two peers disagree about the
+//! shape of an entry, not merely the routes that carry it. The cloud API keys
+//! entries by UUID, timestamps them as RFC 3339 strings, names the harvest
+//! field `source_commit`, and expresses "archived" as a tombstone timestamp
+//! rather than a status word.
```
usercise, line 8: This sounds like cloud api has a more up to date, fully developed version. We should update team server to match rather than create a separate wire type.

I would expect archive to carry an effective time it happened, RFC 3339 is the correct way to carry the date time data over the wire.

usercise: You are right about the substance, and more right than I had assumed. I went and checked rather than defending the current shape.

**The team server does not carry archive time at all.** Its `ServerNote` response schema is `id, kind, title, body, tags, linked_files, created_at, status, superseded_by, remote_id, distance`. There is no `invalid_at`, no `valid_at`, no `archived_at`. So this is not a difference of encoding, it is missing information: the hosted API tells you when an entry was archived and the team server does not.

It is worse than "missing", actually. The client wire type for the team server already declares `valid_at`, `invalid_at` and `source_ref` as `#[serde(default)]` optional fields, so the CLI has been ready to receive them the whole time and the server has never sent them. The domain `Note` has the fields. They are dropped only at the team server's own wire boundary. That is a real defect, and adding them is a strict improvement with a consumer already waiting.

So on the archive-time half: agreed, and it should be fixed.

**On making the team server match cloud-api's encoding, there is a constraint I do not think we should quietly step over.** `docs/stability.md` declares every route under `/v1/` stable, "paths, methods, request and response schemas, and status codes", and then splits it:

- *Allowed:* new routes, new optional request fields, **new response fields**, new enum values in an open field.
- *Not allowed:* removing or renaming a response field, making an optional request field required, **narrowing an accepted type**, changing the meaning of a status code.

Adding `archived_at` / `valid_at` / `invalid_at` is squarely in the allowed column. Re-encoding `created_at` from `integer/int64` to an RFC 3339 string, and `id` from `integer/int64` to a UUID, changes the type of two **required** response fields. That is the not-allowed column, it breaks every already-released CLI against an upgraded server, and the `openapi-snapshot` CI job enforces the committed schema, so it cannot land quietly even by accident.

It also directly contradicts two acceptance criteria of this task, which ask that the self-hosted dialect keep its routes, verbs and payload shapes unchanged.

**On "rather than create a separate wire type":** the honest defence is narrow. The two types are not duplication for its own sake, they encode two independently-versioned public contracts, and merging them means one of those contracts breaks. But the maintenance cost you are pointing at is real, and it does not go away by my calling it a contract.

**What I would do, and what I would like you to pick:**

1. **Now, in a separate small change:** add `valid_at`, `invalid_at` and `source_ref` to the team server's note response. Purely additive, contract-safe, closes the real gap you found, and the CLI already reads them. I would keep this PR's two wire types, because they will still be two encodings.
2. **Separately, as a deliberate decision:** converge the encodings. That is a breaking change to a stable API, so it wants a version bump or a `/v2/`, a compatibility window where the server emits both shapes, and a migration note. Worth doing if we want one wire format, but it is its own piece of work with its own risk, not a tidy-up inside this PR.

I have not started either. Tell me which you want and whether it blocks this PR, and I will pick it up. My instinct is that option 1 should happen soon regardless, and option 2 is a real decision rather than a refactor.

For completeness on RFC 3339 itself: I agree it is the better wire encoding, and it is what this branch already consumes from the hosted API, converting to epoch seconds at the boundary because that is what the domain `Note` and the local SQLite schema use. Converging on RFC 3339 end to end would also touch the local store's representation, which is a third scope again.

---
_Generated by [Claude Code](https://claude.ai/code)_

CHANGELOG.md
```diff
@@ -11,6 +11,41 @@ spelunk uses [Semantic Versioning](https://semver.org/).
...
+
+  Against the hosted API, listing and searching share one route distinguished by
+  a query parameter, entry counts come from the total the list route already
+  computes, archiving is a `DELETE`, and supersession is expressed as a batch
+  edge. Harvest's duplicate check filters client-side, since the hosted API has
+  no `source_commit` filter: correct, but proportional to the size of the
+  project rather than an indexed lookup.
+
+  `project_id` may be a slug or a UUID against either peer for every memory
+  operation, including `spelunk memory show` and `spelunk memory archive`.
```
usercise, line 33: This entry is much too long, please tighten up to a paragraph or less. Suggestion: 

```suggestion
- **`mode = "cloud_first"` now serves memory against the hosted API, not only a
  self-hosted spelunk-server.** 

  `project_id` may be a slug or a UUID against either peer for every memory
  operation, including `spelunk memory show` and `spelunk memory archive`.
```

crates/spelunk-core/src/storage/remote/cloud_api/wire.rs
usercise: Applied in `92856ac`, your suggestion as written.

Nothing is lost by cutting it: the dialect-detection rule and the client-side harvest filtering cost both already live in the `docs/memory.md` section this PR adds, which is where someone actually looks them up. The changelog only needed to say the capability exists and how `project_id` behaves.

Only the CHANGELOG changed, so no code gates were re-run; the diff hygiene checks (tracker references, em-dashes, trailing whitespace) are clean.

One thing I did **not** touch, since you flagged one entry and I would rather ask than trim your prose uninvited: the `### Changed` entry just below it, on ids becoming opaque, is nine lines and the same shape. If it reads long to you too, the second paragraph about `--format json` output is the load-bearing half, since it is the only part that tells an existing user whether their scripts break. Say the word and I will cut it to that.

---
_Generated by [Claude Code](https://claude.ai/code)_

CHANGELOG.md
usercise: Applied in `92856ac`, your suggestion as written.

Nothing is lost by cutting it: the dialect-detection rule and the client-side harvest filtering cost both already live in the `docs/memory.md` section this PR adds, which is where someone looks them up. The changelog only needed to say the capability exists and how `project_id` behaves.

Only the CHANGELOG changed, so no code gates were re-run. The diff hygiene checks (tracker references, em-dashes, trailing whitespace) are clean.

One thing I did **not** touch, since you flagged one entry and I would rather ask than trim prose uninvited: the `### Changed` entry just below, on ids becoming opaque, is nine lines and the same shape. If it reads long to you too, the `--format json` paragraph is the load-bearing half, since it is the only part that tells an existing user whether their scripts break. Say the word and I will cut it to that.

(Apologies for the noise: I posted this reply in the `wire.rs` thread first by mistake. Ignore that copy, this is the right thread.)

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #812, opened 2026-08-04
docs(adr): import notes carrier into memory.db on read

usercise:
## What

Adds ADR-077, completing ADR-069. ADR-069 D5 merges the fetched tracking ref (`refs/notes/origin/spelunk`) into the working ref (`refs/notes/spelunk`) on read paths, but the default read paths query the SQLite `memory.db`, which nothing imports the carrier into except `init`. So a teammate's published memory stays invisible on the default path until a manual re-`init`.

## Decision

- **Read paths import the carrier into `memory.db`**, gated on a notes-ref OID marker persisted in `memory.db`, so the steady-state hot path spawns no git subprocess and runs no note-blob walk. Import-every-read (too expensive on a CLI shelled out thousands of times/session) and an explicit `sync` command (reintroduces the invisible-until-manual-step failure) were both rejected.
- **`init` configures the refspec and fetches notes *before* its import pass**, so one `init` after clone suffices instead of `clone → init → fetch → init again`.
- **No new git hook** — git has no post-fetch hook and `post-merge` misses bare `git fetch`; the read-path import covers every consumer with nothing to install. Publishing stays on the existing opt-in pre-push hook.
- **`init` stages `.spelunk/config.toml`** (does not commit), making the documented committed-carrier promise actionable for slugs that cannot be re-derived on a fresh clone.

## Acceptance guard

The prior work in this area was a false-close: the merge was asserted via `spelunk memory --backend git-notes list`, which reads the git ref directly, not the default `memory.db`. The acceptance criteria here require a two-clone round trip that reads the **default** SQLite path end to end (clone A publishes → clone B fetches → clone B's default read surfaces it, no manual re-init).

Design only — no `crates/` changes. The ADR carries an engineer-ready implementation contract (acceptance criteria + Canon-TDD test list).

### DIFF
--- a/docs/adr/077-teammate-memory-reaches-read-paths.md
+++ b/docs/adr/077-teammate-memory-reaches-read-paths.md
@@ -1,6 +1,5 @@
 # ADR-077: Teammate memory reaches read paths — import the notes carrier into `memory.db` on read, gated by a notes-ref OID marker
 
-**Status:** Proposed
 **Date:** 2026-08-04
 **Deciders:** founder (Johan); architect
 **Relationship to prior ADRs:** completes
--- a/docs/adr/077-teammate-memory-reaches-read-paths.md
+++ b/docs/adr/077-teammate-memory-reaches-read-paths.md
@@ -1,0 +1,336 @@
+# ADR-077: Teammate memory reaches read paths — import the notes carrier into `memory.db` on read, gated by a notes-ref OID marker
+
+**Status:** Proposed
+**Date:** 2026-08-04
+**Deciders:** founder (Johan); architect
+**Relationship to prior ADRs:** completes
+[ADR-069](069-git-notes-sharing-pre-push-hook-and-tracking-refspec.md). ADR-069
+D5 decided that spelunk merges the fetched tracking ref
+(`refs/notes/origin/spelunk`) into the working ref (`refs/notes/spelunk`) on its
+own read paths, so "reading a teammate's memory needs no opt-in." That merge is
+implemented and tested, but it stops at the git ref. The default read paths query
+the SQLite `memory.db`, which nothing imports the carrier into except `init`, so
+the merged entries never actually surface. This ADR carries the notes across the
+last gap — ref to `memory.db` — on every read, cheaply. Leaves ADR-068's carrier
+model and ADR-004's inference-vs-storage split untouched: the carrier is still
+`refs/notes/spelunk` and the queryable store is still `memory.db`; this ADR only
+keeps the second in step with the first.
+
+## Context
+
+Two reproducible failures leave a teammate's published memory invisible.
+
+### Failure 1 — a fresh clone imports zero memory
+
+`git clone` does not fetch `refs/notes/*`. `spelunk init`
+(`crates/spelunk-cli/src/cli/cmd/init.rs`) runs its git-notes import pass
+(step 6b, `import_git_notes_into_memory`) **before** it configures the `origin`
+notes fetch refspec (step 7, `configure_notes_refspec`). On a fresh clone the
+tracking ref has never been fetched — the refspec that would fetch it is set only
+in step 7, which runs after the import — so step 6b's `merge_tracking_notes` +
+`import_git_notes_into_memory` run against an absent/empty notes ref and import
+nothing. The only sequence that currently works is
+`clone → init → git fetch → init again`: the first `init` sets the refspec, the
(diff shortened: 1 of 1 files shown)


docs/adr/077-teammate-memory-reaches-read-paths.md
```diff
@@ -0,0 +1,336 @@
+# ADR-077: Teammate memory reaches read paths — import the notes carrier into `memory.db` on read, gated by a notes-ref OID marker
+
+**Status:** Proposed
```
usercise: Status can't be proposed if this PR is ready for review.

usercise: Fixed in the latest push — removed the "Status: Proposed" line. Recent ADRs (070, 076) carry no Status field, and per house convention approval is the merge, so the line was both contradictory for a ready PR and off-style. Header otherwise unchanged (Date / Deciders / Relationship to prior ADRs).

docs/adr/077-teammate-memory-reaches-read-paths.md
```diff
@@ -0,0 +1,335 @@
...
+reviewer — with nothing to install, and it is idempotent and cheap via the D2
+gate, so a hook would be both redundant with the read path and incomplete
+relative to it. Publishing memory stays on the existing opt-in **pre-push** hook
+(ADR-069 D1/D7), unchanged.
+
+### D5 — `init` stages `.spelunk/config.toml`, and does not commit it
+
+`init` `git add`s `.spelunk/config.toml` after writing the slug, best-effort and
+non-fatal like the rest of `init`. It does **not** commit.
+
```
usercise: This feels like an over-reach by the tool. If we're already saying we don't have authority to commit the change then I don't think we should claim authority to stage things either. This is the kind of change that will annoy someone because they didn't see notice it in the status before they committed. In short no `git add` from `init`.

docs/adr/077-teammate-memory-reaches-read-paths.md
```diff
@@ -0,0 +1,335 @@
...
+- The `memory add` secret scan (`contains_secret`, run before any persistence) is
+  unchanged and still runs on the write path into both `memory.db` and the note,
+  so imported content was already scanned at its origin.
+- The D2 marker holds only a git object id, no secret.
+
+## Implementation contract
+
+> The architect does not write the Rust or the tests. This section is the
+> contract the implementer drives off, and it is reproduced in the handoff report.
+
```
usercise: Why did this go in the ADR? This is an agent instruction.

docs/adr/077-teammate-memory-reaches-read-paths.md
```diff
@@ -0,0 +1,335 @@
...
+  no new fetch does not import again.
+- **Offline `init` still succeeds.** With no reachable `origin`, `init` completes
+  (D3 fetch is best-effort) and configures the refspec.
+- **`init` stages `config.toml`.** After `init` in a repo, `.spelunk/config.toml`
+  is staged (present in `git diff --cached --name-only`) and no commit was created
+  by `init`.
+- **No regression to local-only or cloud-routing reads.** With no git repo, or on
+  a `cloud_first` config, reads behave exactly as today (no import attempt).
+
+### Canon-TDD test list
```
usercise: Not in an ADR.

docs/adr/077-teammate-memory-reaches-read-paths.md
usercise: Reversed D5 (commit d33b29d) — init now performs no git operation on .spelunk/config.toml: no git add, no commit. It writes the file; the docs (getting-started + commands) now phrase committing it as an explicit user step, and init may print a one-line advisory. Rationale matches yours: if the tool lacks authority to commit, it lacks authority to silently mutate the index either.

docs/adr/077-teammate-memory-reaches-read-paths.md
usercise: Agreed (commit d33b29d) — the entire Implementation contract section, including that agent-instruction blockquote, is removed. The ADR now holds only decision, context/rationale, consequences, non-goals, and security.

docs/adr/077-teammate-memory-reaches-read-paths.md
usercise: Removed (commit d33b29d) — the acceptance criteria and Canon-TDD test list are gone from the ADR (they live on the board task instead). Also dropped the one Consequences bullet that duplicated that test directive.

PR #815, opened 2026-08-04
fix(logout): scope --server/--servers to server keys only

usercise:
## The bug

`spelunk logout --server <url>` clears the named per-origin self-hosted server key **and also** deletes the spelunk.cloud credential pair (the `[auth]` block: `access_token`/`refresh_token`/`expires_at`/`org_id`). A user clearing one self-hosted key is silently signed out of spelunk.cloud and forced to redo a browser device login. `--servers` had the same side effect.

Root cause: in `crates/spelunk-cli/src/cli/cmd/logout.rs`, `config::remove_auth_tokens()` ran unconditionally at the top of `logout()`, before the flag check — so every form wiped the cloud pair.

## The fix

Move the cloud-pair clear into the bare, no-flag branch. Each `logout` form now touches exactly one credential store:

- `spelunk logout` (no flags): clears **only** the cloud `[auth]` token pair; all server keys survive.
- `spelunk logout --server <url>`: clears **only** that one origin's server key; cloud pair and other origins survive.
- `spelunk logout --servers`: clears **only** all per-origin server keys (map + legacy entry); cloud pair survives.

Storage model, for reference: the cloud token pair lives in the `[auth]` table of `~/.config/spelunk/config.toml`; per-origin self-hosted keys live in the secret store under the `server_keys` map (with a legacy flat `server_key` fallback).

Also tightened the now-misleading `--server`/`--servers` help text (dropped "Also clear…", which implied the cloud pair was cleared too) and the module doc.

## Tests

New `crates/spelunk-cli/tests/logout_credential_scope.rs` — three tests, one per form, each seeding **both** stores (a full `[auth]` pair + two server keys) and asserting which store changed and which survived:

- `logout_server_flag_clears_that_origin_only_and_keeps_cloud_pair`
- `bare_logout_clears_cloud_pair_only_and_keeps_server_keys`
- `logout_servers_flag_clears_all_server_keys_and_keeps_cloud_pair`

The two flag-form tests were red before the fix (cloud pair wiped) and are green after. The existing `auth_server_keys.rs` logout tests still pass — they only ever asserted the server-key side, which is why the regression slipped through.

Run:

```
SPELUNK_SECRET_STORE=file cargo test -p spelunk-cli --test logout_credential_scope --test auth_server_keys
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,18 @@ spelunk uses [Semantic Versioning](https://semver.org/).
   upgrade the CLI to match. See `docs/version-skew.md`. The `GET
   /memory/since?t=` legacy mode is unchanged, and still returns a bare array.
 
+### Fixed
+
+- **`spelunk logout --server <url>` no longer signs you out of spelunk.cloud.**
+  It was clearing the cloud token pair (the `[auth]` block written by `spelunk
+  login`) as a side effect of clearing the one named self-hosted server key,
+  forcing an unnecessary browser device login to get back in. `--server <url>`
+  now clears only that one origin's key, `--servers` clears only every stored
+  server key, and both leave the cloud token pair intact; bare `spelunk logout`
+  still clears only the cloud pair. Each form now touches exactly one
+  credential store. The `--server`/`--servers` help text no longer says
+  "Also clear…", which implied the cloud pair was cleared too.
+
 ## [0.9.6] — 2026-07-31
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,14 +26,6 @@ spelunk uses [Semantic Versioning](https://semver.org/).
 ### Fixed
 
 - **`spelunk logout --server <url>` no longer signs you out of spelunk.cloud.**
-  It was clearing the cloud token pair (the `[auth]` block written by `spelunk
-  login`) as a side effect of clearing the one named self-hosted server key,
-  forcing an unnecessary browser device login to get back in. `--server <url>`
-  now clears only that one origin's key, `--servers` clears only every stored
-  server key, and both leave the cloud token pair intact; bare `spelunk logout`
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -25,6 +25,15 @@ spelunk uses [Semantic Versioning](https://semver.org/).
...
 
+- **`spelunk logout --server <url>` no longer signs you out of spelunk.cloud.**
+  It was clearing the cloud token pair (the `[auth]` block written by `spelunk
+  login`) as a side effect of clearing the one named self-hosted server key,
+  forcing an unnecessary browser device login to get back in. `--server <url>`
+  now clears only that one origin's key, `--servers` clears only every stored
+  server key, and both leave the cloud token pair intact; bare `spelunk logout`
+  still clears only the cloud pair. Each form now touches exactly one
+  credential store. The `--server`/`--servers` help text no longer says
+  "Also clear…", which implied the cloud pair was cleared too.
```
```suggestion
- **`spelunk logout --server <url>` no longer signs you out of spelunk.cloud.**
```

This entry is too long for a changelog. The important information was all in the first sentence.

crates/spelunk-cli/src/cli/cmd/logout.rs
```diff
@@ -6,7 +6,10 @@
 //! founder-review correction): a developer recovering from a broken cloud
 //! login should not silently lose the server key(s) they use on other
 //! projects. Clearing those is an explicit, separate action via `--servers`
-//! (all of them) or `--server <url>` (just one).
+//! (all of them) or `--server <url>` (just one). The scoping is symmetric:
+//! those two forms clear **only** server keys and leave the cloud `[auth]`
+//! pair untouched, so clearing one self-hosted key never signs the developer
+//! out of spelunk.cloud. Each form touches exactly one credential store.
```
usercise:
```suggestion
//! (all of them) or `--server <url>` (just one).
```

The comment was fine as it was.

PR #816, opened 2026-08-04
fix(memory): reject unknown `memory add --kind` instead of silently storing it

usercise:
## The bug

`spelunk memory add --kind <anything>` accepted any string as the kind, printed `Stored [...]`, and exited 0:

```
$ spelunk memory add --kind decisions --title x --body y
Stored [decisions] #11: x        # exit 0
```

Because every retrieval path filters on an exact kind — `memory list --kind decision`, `spelunk context` (which selects handoffs/questions/decisions/requirements by kind), and `memory failures` (which selects `antipattern`) — a typo like `--kind decisions` or `--kind desicion` stored an entry that none of them could ever surface, silently dropping agent-recorded decisions out of every retrieval path while reporting success.

## The fix

Introduce a single canonical source of truth in `spelunk-core`, `storage::note_kind`:

- `NOTE_KINDS` — the nine valid kinds (`decision`, `context`, `requirement`, `note`, `question`, `answer`, `handoff`, `intent`, `antipattern`).
- `is_valid_note_kind` / `parse_note_kind` — membership check and a strict parser whose error names the offending value and lists the valid kinds.

`parse_note_kind` is wired as the clap `value_parser` for `memory add --kind`, so an unknown kind is rejected **at argument-parse time, before any store is opened**, with a non-zero exit and no write:

```
$ spelunk memory add --kind decisions --title x --body y
error: invalid value 'decisions' for '--kind <KIND>': unknown kind 'decisions' — valid kinds are: decision, context, requirement, note, question, answer, handoff, intent, antipattern
# exit 2
```

The default (`note`) and all nine valid kinds are unchanged.

## Where validation lives / drift

`memory add --kind` is the only place user-supplied kinds enter, so validation is enforced there (the clap value parser), reading from the one canonical `NOTE_KINDS` list. Retrieval paths deliberately select on a *subset* of those kinds, so there is no pre-existing "retrieval canon" to assert equality against; instead a guard test asserts every kind `context` and `memory failures` select on is a member of `NOTE_KINDS`, so validation and retrieval can't silently diverge (a retrieval kind can never be one `memory add` would reject). Legacy/remote/harvested entries are not revalidated — only user-supplied `--kind` on `memory add` is gated.

## Tests

- `spelunk-core` unit (`storage::note_kind::tests`): canonical set is exactly the nine kinds; every canonical kind is valid; unknown/typo kinds are invalid; `parse_note_kind` accepts each valid kind unchanged and rejects unknowns with a message naming the value and listing the kinds.
- CLI e2e (`tests/memory_add_kind_validation.rs`): each of the nine kinds is accepted and stored; omitting `--kind` defaults to `note`; `--kind bogus` and realistic typos (`decisions`, `desicion`) are rejected with a non-zero exit, a message naming the value and listing the kinds, and **no** row written.
- CLI guard (`context::tests::every_retrieval_kind_is_a_canonical_kind`): every retrieval kind is a canonical kind.

New tests fail before the fix (an unknown kind is accepted and stored) and pass after; the existing memory/context/help-text suites remain green with no regressions.

## Follow-up (out of scope)

No opt-in/extensibility escape hatch was added, per the task scope. If teams later need project-specific kinds, a deliberate config-driven extension of `NOTE_KINDS` would be the place to add it — flagged here rather than built.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,13 +27,7 @@ spelunk uses [Semantic Versioning](https://semver.org/).
 
 - **`spelunk memory add --kind` now rejects an unknown kind instead of silently
   storing it.** Previously any string was accepted, so a typo (`--kind
-  decisions`, `--kind desicion`) stored an entry that no retrieval path —
-  `memory list --kind decision`, `spelunk context`, `memory failures` — could
-  ever surface, even though the command printed `Stored [...]` and exited 0.
-  `memory add` now validates `--kind` against the nine canonical kinds
-  (`decision`, `context`, `requirement`, `note`, `question`, `answer`,
-  `handoff`, `intent`, `antipattern`) and exits non-zero with a message naming
-  the offending value and listing the valid kinds; nothing is written. The
+  decisions`, `--kind desicion`) stored an entry with no retrieval path. The
   default (`note`) and every valid kind are unaffected.
 
 ## [0.9.6] — 2026-07-31
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,19 @@ spelunk uses [Semantic Versioning](https://semver.org/).
   upgrade the CLI to match. See `docs/version-skew.md`. The `GET
   /memory/since?t=` legacy mode is unchanged, and still returns a bare array.
 
+### Fixed
+
+- **`spelunk memory add --kind` now rejects an unknown kind instead of silently
+  storing it.** Previously any string was accepted, so a typo (`--kind
+  decisions`, `--kind desicion`) stored an entry that no retrieval path —
+  `memory list --kind decision`, `spelunk context`, `memory failures` — could
+  ever surface, even though the command printed `Stored [...]` and exited 0.
+  `memory add` now validates `--kind` against the nine canonical kinds
+  (`decision`, `context`, `requirement`, `note`, `question`, `answer`,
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -23,6 +23,19 @@ spelunk uses [Semantic Versioning](https://semver.org/).
...
+- **`spelunk memory add --kind` now rejects an unknown kind instead of silently
+  storing it.** Previously any string was accepted, so a typo (`--kind
+  decisions`, `--kind desicion`) stored an entry that no retrieval path —
+  `memory list --kind decision`, `spelunk context`, `memory failures` — could
+  ever surface, even though the command printed `Stored [...]` and exited 0.
+  `memory add` now validates `--kind` against the nine canonical kinds
+  (`decision`, `context`, `requirement`, `note`, `question`, `answer`,
+  `handoff`, `intent`, `antipattern`) and exits non-zero with a message naming
+  the offending value and listing the valid kinds; nothing is written. The
+  default (`note`) and every valid kind are unaffected.
```
usercise, line 31:
```suggestion
- **`spelunk memory add --kind` now rejects an unknown kind instead of silently
  storing it.** Previously any string was accepted, so a typo (`--kind
  decisions`, `--kind desicion`) stored an entry with no retrieval path. The
  default (`note`) and every valid kind are unaffected.
```

PR #819, opened 2026-08-04
fix(search): gate explicit --mode semantic/hybrid when no server reachable

usercise:
## The bug

With no server reachable, an explicit `spelunk search --mode semantic` (and `--mode hybrid`) failed **silently** instead of with the actionable locked-feature error every other inference-gated command gives:

```
SPELUNK_NO_SERVER=1 spelunk search "x" --mode semantic
#   No results found.
#   exit 0
```

To an agent or script, `No results found.` + exit 0 reads as "no such code exists", not "the feature is unavailable". Every sibling command (`memory search`, `explore`, `memory timeline`, `plumbing embed`) instead fails closed with `'... requires spelunk-server. Run \`spelunk server start\` ...` and a non-zero exit.

Only the *explicit* `semantic`/`hybrid` path was missing the gate. The default `auto` mode already degrades correctly (prints a notice, falls back to ast-grep).

## The fix

The explicit-mode block in `crates/spelunk-cli/src/cli/cmd/search.rs` used to silently switch to FTS text search when no server was reachable. It now routes those modes through the shared `require_server_client(&cfg, "search")` helper — the exact gate `memory search` / `memory timeline` / `plumbing embed` use — so the message wording and exit code stay consistent (no hand-rolled string). This covers both the `SPELUNK_NO_SERVER=1` kill-switch and the plain no-server-reachable case (both are Tier 0).

`auto` is deliberately untouched: it still announces its degradation and falls back to ast-grep with exit 0.

## Tests

New `crates/spelunk-cli/tests/search_semantic_no_server_gate.rs` (drives the real CLI against a populated-but-unembedded offline index):

- `search --mode semantic` with no server → exit non-zero + `requires spelunk-server` / `spelunk server start` on stderr; never `No results found.`
- `search --mode hybrid` with no server → same gate.
- `search --mode auto` with no server → unchanged: degrades to ast-grep, exit 0.

Verified the semantic/hybrid tests were **red before** the fix (`code=0`, `stdout="No results found."`) and green after. No regressions in `search_empty_index_mode`, `inference_server_message`, `egress_containment` (incl. server-backed `search_semantic_zero_egress`), or the `search` module unit tests. `cargo fmt --all -- --check` and `cargo clippy -p spelunk-cli --tests` are clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,19 @@ spelunk uses [Semantic Versioning](https://semver.org/).
   upgrade the CLI to match. See `docs/version-skew.md`. The `GET
   /memory/since?t=` legacy mode is unchanged, and still returns a bare array.
 
+### Fixed
+
+- **`spelunk search --mode semantic` (and `--mode hybrid`) now fail with an
+  actionable error when no server is reachable, instead of silently reporting
+  "No results found." and exiting 0.** These modes need a server to embed the
+  query, so with none reachable (including under `SPELUNK_NO_SERVER=1`) they now
+  emit the same locked-feature error as the other inference-backed commands
+  (`'spelunk search' requires spelunk-server. Run `spelunk server start` ...`)
+  and exit non-zero. Previously the empty fallback result read to an agent or
+  script as "this code does not exist" rather than "the feature is unavailable".
+  The default `auto` mode is unchanged: it still announces its degradation and
+  falls back to ast-grep.
+
 ## [0.9.6] — 2026-07-31
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,10 +29,7 @@ spelunk uses [Semantic Versioning](https://semver.org/).
   actionable error when no server is reachable, instead of silently reporting
   "No results found." and exiting 0.** These modes need a server to embed the
   query, so with none reachable (including under `SPELUNK_NO_SERVER=1`) they now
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -25,21 +25,16 @@ spelunk uses [Semantic Versioning](https://semver.org/).
...
+- **`spelunk search --mode semantic` (and `--mode hybrid`) now fail with an
+  actionable error when no server is reachable, instead of silently reporting
+  "No results found." and exiting 0.** These modes need a server to embed the
+  query, so with none reachable (including under `SPELUNK_NO_SERVER=1`) they now
+  emit the same locked-feature error as the other inference-backed commands
+  (`'spelunk search' requires spelunk-server. Run `spelunk server start` ...`)
+  and exit non-zero. Previously the empty fallback result read to an agent or
+  script as "this code does not exist" rather than "the feature is unavailable".
+  The default `auto` mode is unchanged: it still announces its degradation and
+  falls back to ast-grep.
```
usercise, line 34:
```suggestion
- **`spelunk search --mode semantic` (and `--mode hybrid`) now fail with an
  actionable error when no server is reachable, instead of silently reporting
  "No results found." and exiting 0.** These modes need a server to embed the
  query, so with none reachable (including under `SPELUNK_NO_SERVER=1`) they now
  emit the same locked-feature error as the other inference-backed commands.
  The default `auto` mode is unchanged: it still announces its degradation and
  falls back to ast-grep.
```

PR #836, opened 2026-08-06
Add a standalone tool that exports local stores to a portable dump

usercise:
A small, self-contained binary that reads a project's local stores and writes them as a portable, documented dump. Useful for taking a backup, inspecting what is actually stored, moving a project between machines, or leaving the tool entirely with your data in a form something else can read.

**Status: ready for review. Two independent reviews completed; all blocking findings fixed.** CI is green across all 15 checks.

> Both reviews were posted as comments rather than formal reviews, so GitHub shows no review decision on this PR. That is a tooling artifact, not an indication that it is unreviewed — see the review threads below.

## What it does

Reads the memory store, the project registry and the server store, and writes line-delimited JSON: a header, then entities, then relationships, then a footer.

The format is expressed as **entities and relationships rather than tables**, so it describes what is stored rather than how this version happens to store it. A reader does not need to know our schema, and the format does not have to change when the schema does.

## Properties worth knowing

**It never modifies anything.** Read-only against every store. The tests assert the sources are byte-identical afterwards.

**It writes to a temporary file, verifies, then renames.** A failed run leaves no dump at all, rather than a truncated one that looks complete.

**The footer carries per-type counts and a digest** folded from per-record hashes in file order. That gives per-record integrity, order sensitivity, and whole-or-nothing verification in one mechanism, so a truncated or reordered dump fails loudly instead of importing partially.

**No version constant anywhere.** It probes for columns rather than branching on a recorded schema version, so one code path reads every shape the stores have had, including older ones that predate version stamping. That is deliberate: a version-keyed reader would need updating every time the schema moves, and would fail on any store whose version was never written.

**It carries identifiers verbatim and never mints them.** Where a source has an identifier it is copied; where it does not, the field is omitted and the creation time is always carried so a reader can decide. Identity policy belongs to whatever consumes the dump, not to the exporter.

**It links its own SQLite** and depends on no other crate in the workspace, so it builds and runs standalone.

## Testing

42 tests across the format, the sources and end-to-end runs. Beyond fixtures, it was exercised against real data on a working machine: a live memory store and a registry holding several hundred projects, both self-verified, with the sources unchanged afterwards.

Fixture coverage spans twenty store shapes built by applying the schema history step by step, plus stores carrying no version stamp at all.

## Not included

Scanning for hooks, environment variables and lock files. The module is present but unimplemented: nothing in this change exercises it, and building it now would be structure without a caller.

## Review notes

The two places I would look hardest:

- **`source.rs`** does the column probing. The question worth asking is whether any store shape is read as a *different* shape rather than failing, since a systematic misread is the one error the self-verification cannot catch by construction.
- **Supersede relationships have two encodings in the stores, and one store can hold both.** They point opposite ways relative to each other. A single-encoding fixture will not catch a writer that gets the direction wrong, so that is the sharpest trap in here.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,31 @@ spelunk uses [Semantic Versioning](https://semver.org/).
 
 ## [Unreleased]
 
+### Added
+
+- **`spelunk-export`, a standalone tool that writes a portable dump of a local
+  store.** Everything the product records lives in local SQLite databases with
+  no supported way to get it out, so backing it up, inspecting it, moving it to
+  another machine or leaving the product all required writing SQL by hand
+  against a schema that has changed ten times. The dump is line-delimited JSON,
+  one record per line, expressed as entities and relationships rather than as a
+  copy of any table, and readable without special tooling. Stores are opened
+  read-only and are never modified. Only authored data is carried: full-text
+  indexes, embeddings and import cursors all regenerate, so carrying them would
+  only risk carrying stale state forward. The tool makes no network request and
+  needs no SQLite installed on the host. An `inventory` subcommand reports what
+  a store holds without touching it.
+
+  Every table is read inside a single read transaction, so a dump is a
+  consistent point-in-time view of the whole store. Reading each table
+  separately would let a writer that commits an entry and then a link to it land
+  between the two reads, producing a link that appears to point at nothing:
+  live data that would be dropped, blamed on a broken store, and certified as a
+  clean export. A link whose endpoint really is missing from the store, which
+  data written before foreign keys were enforced can still contain, is reported
+  and left out, and the run states how many links that was alongside what it
+  carried.
+
 ## [0.9.7] — 2026-08-05
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,33 +12,15 @@ spelunk uses [Semantic Versioning](https://semver.org/).
 ### Added
 
 - **`spelunk-export`, a standalone tool that writes a portable dump of a local
-  store.** Everything the product records lives in local SQLite databases with
-  no supported way to get it out, so backing it up, inspecting it, moving it to
-  another machine or leaving the product all required writing SQL by hand
-  against a schema that has changed ten times. The dump is line-delimited JSON,
-  one record per line, expressed as entities and relationships rather than as a
-  copy of any table, and readable without special tooling. Stores are opened
-  read-only and are never modified. Only authored data is carried: full-text
-  indexes, embeddings and import cursors all regenerate, so carrying them would
-  only risk carrying stale state forward. The tool makes no network request and
-  needs no SQLite installed on the host. An `inventory` subcommand reports what
-  a store holds without touching it.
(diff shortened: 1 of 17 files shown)


CHANGELOG.md
```diff
@@ -9,6 +9,37 @@ spelunk uses [Semantic Versioning](https://semver.org/).
...
+  no supported way to get it out, so backing it up, inspecting it, moving it to
+  another machine or leaving the product all required writing SQL by hand
+  against a schema that has changed ten times. The dump is line-delimited JSON,
+  one record per line, expressed as entities and relationships rather than as a
+  copy of any table, and readable without special tooling. Stores are opened
+  read-only and are never modified. Only authored data is carried: full-text
+  indexes, embeddings and import cursors all regenerate, so carrying them would
+  only risk carrying stale state forward. The tool makes no network request and
+  needs no SQLite installed on the host. An `inventory` subcommand reports what
+  a store holds without touching it.
```
```suggestion
- **`spelunk-export`, a standalone tool that writes a portable dump of a local
  store.** The dump is line-delimited JSON,
  one record per line, expressed as entities and relationships rather than as a
  copy of any table, and readable without special tooling. Stores are opened
  read-only and are never modified. Only authored data is carried: full-text
  indexes, embeddings and import cursors all regenerate, so carrying them would
  only risk carrying stale state forward. Every table is read inside a single read 
  transaction, so a dump is a consistent point-in-time view of the whole store.
  
  An `inventory` subcommand reports what a store holds without touching it.
```

CHANGELOG.md
```diff
@@ -9,6 +9,37 @@ spelunk uses [Semantic Versioning](https://semver.org/).
...
+  Every table is read inside a single read transaction, so a dump is a
+  consistent point-in-time view of the whole store. Reading each table
+  separately would let a writer that commits an entry and then a link to it land
+  between the two reads, producing a link that appears to point at nothing:
+  live data that would be dropped, blamed on a broken store, and certified as a
+  clean export. A link whose endpoint really is missing from the store, which
+  data written before foreign keys were enforced can still contain, is reported
+  and left out, and the run states how many links that was alongside what it
+  carried.
+
```
```suggestion
```

CHANGELOG.md
```diff
@@ -9,6 +9,37 @@ spelunk uses [Semantic Versioning](https://semver.org/).
...
+- **`spelunk-export`, a standalone tool that writes a portable dump of a local
+  store.** Everything the product records lives in local SQLite databases with
+  no supported way to get it out, so backing it up, inspecting it, moving it to
+  another machine or leaving the product all required writing SQL by hand
+  against a schema that has changed ten times. The dump is line-delimited JSON,
+  one record per line, expressed as entities and relationships rather than as a
+  copy of any table, and readable without special tooling. Stores are opened
+  read-only and are never modified. Only authored data is carried: full-text
+  indexes, embeddings and import cursors all regenerate, so carrying them would
```
usercise: Dropping `embeddings` and regenerating them is a blocker for me. That would mean that on a fresh import we would need to ask people to run possibly a multi-hour to multi-day re-index, vs carrying them and asking them to do a few minutes migration & re-index.

PR #846, opened 2026-08-10
ci(release): publish spelunk-export as a standalone per-platform asset

usercise:
## Why

`crates/spelunk-export` merged, but `.github/workflows/release.yml` never
mentioned it. Tagging a release today builds the binary and throws it away, so
there is no way for anyone to obtain the tool.

## What changed

`spelunk-export` is now packaged and uploaded as **its own asset per
platform**, for the four targets already in the release matrix:

```
spelunk-export-<tag>-x86_64-unknown-linux-gnu.tar.gz
spelunk-export-<tag>-aarch64-unknown-linux-gnu.tar.gz
spelunk-export-<tag>-aarch64-apple-darwin.tar.gz
spelunk-export-<tag>-x86_64-pc-windows-msvc.zip
```

It is deliberately **not** in the product archive, the `.deb`, the Homebrew
formula or the Scoop manifest. The tool is meant to be fetched on its own,
verified, run against a store and deleted again; making that require a download
of the whole product, or putting a read-only tool permanently on `PATH`, would
defeat the point.

No new pipeline: the workspace build already produced the binary, so this is
packaging plus coverage.

- The glibc-2.31 ceiling check now covers all three binaries, not two. An asset
  that cannot run on the support floor fails the release instead of shipping.
- `tarball-smoke` downloads the new archive and, in the same `debian:11` /
  `ubuntu:20.04` / `ubuntu:24.04` / `debian:12` containers, runs it against the
  store the existing memory round trip just wrote. `--version` would not have
  been evidence; reading a real store is what proves the statically linked
  SQLite works on an image carrying no `libsqlite3` at all.
- `scripts/release-dry-run.sh` packages the asset the same way, asserts with
  `ldd` that it links neither a system SQLite nor `libdbus`, and exercises it in
  the `.deb` smoke containers.
- `docs/releasing.md` gains the asset in the pipeline description and the
  download-URL patterns, plus how to read an asset's sha256 from the release.

## Checksums: nothing added, and why

**GitHub's per-asset digest already satisfies this.** Every release asset
carries a `digest` field shaped `sha256:<hex>`, recorded at upload time and
readable per asset without downloading it:

```bash
gh release view <tag> --json assets --jq '.assets[] | "\(.digest)  \(.name)"'
```

This workflow already depends on exactly that in `update-homebrew-formula` and
`update-scoop-manifest`, which is empirical proof the field is populated for
this repo's assets rather than a documentation claim. A generated `SHASUMS`
file would restate a platform guarantee in a form the platform does not verify,
and would need its own consistency check. Nothing was added.

One distinction worth stating plainly: the digest attests that a download
matches what GitHub stored, not that it matches what the build produced. That
is what fetch-then-verify needs. Build provenance would be a different
mechanism and is not claimed here.

## Dynamic linkage: verified, not assumed

`rusqlite` is inherited from `[workspace.dependencies]` with
`features = ["bundled"]`, and `spelunk-export` depends on no other crate in the
workspace (`anyhow`, `clap`, `rusqlite`, `serde`, `serde_json`, `sha2`).

Built inside `debian:11`, `ldd` reports only:

```
libgcc_s.so.1  libpthread.so.0  libm.so.6  libdl.so.2  libc.so.6
```

No `libsqlite3`, and no `libdbus` (which the product binaries do link, for the
keyring). On `aarch64-apple-darwin`, `otool -L` reports only
`/usr/lib/libSystem.B.dylib` and `/usr/lib/libiconv.2.dylib`.

Note the wording: SQLite is statically linked, but the binary is a normal
dynamically linked executable against the platform libc. It is not a fully
static (musl-style) artifact, and the glibc-2.31 floor applies to it exactly as
it does to the other two.

## Dry run

`scripts/release-dry-run.sh` passes end to end:

```
=== release-dry-run: build (debian:11, target x86_64-unknown-linux-gnu) ===
=== release-dry-run: glibc ceiling check (floor: GLIBC_2.31) ===
    spelunk:        max versioned glibc symbol = GLIBC_2.30
    spelunk-server: max versioned glibc symbol = GLIBC_2.30
    spelunk-export: max versioned glibc symbol = GLIBC_2.30
=== release-dry-run: assemble .deb layout + derive Depends (debian:11) ===
=== release-dry-run: build .deb (-Zxz) ===
=== release-dry-run: package export tool archive + check dynamic linkage ===
=== release-dry-run: install + smoke-test .deb and export tool (debian:11) ===
=== release-dry-run: install + smoke-test .deb and export tool (ubuntu:20.04) ===
=== release-dry-run: install + smoke-test .deb and export tool (ubuntu:24.04) ===
=== release-dry-run: PASS ===
```

In each of the three images the tool read the store the smoke test had just
written and verified its own output:

```
[ { "path": "/w/.spelunk/memory.db", "exists": true, "bytes": 106496,
    "schema_version": 10,
    "contents": [ { "table": "notes", "rows": 1 },
                  { "table": "memory_edges", "rows": 0 } ] } ]
Wrote /tmp/dump.jsonl
  1 memory_entry
  the dump reads back as exactly what was read from the store
```

Produced `spelunk-export-<version>-x86_64-unknown-linux-gnu.tar.gz` (1.4 MB).

### What the dry run does not cover

It exercises the Linux **x86_64** leg only. It does **not** exercise:

- the macOS (`aarch64-apple-darwin`) build or packaging step;
- the Windows (`x86_64-pc-windows-msvc`) build, or the `Compress-Archive` step
  that produces the `.zip`;
- the **arm64 Linux** leg (`aarch64-unknown-linux-gnu`);
- the GitHub Release itself, so neither `gh release create`'s globs picking up
  the new archives nor the `digest` field on the new assets is observed here;
- the `tarball-smoke` job as a job, including the `download-artifact` /
  `upload-artifact` round trip;
- the Homebrew or Scoop publish steps.

Those are first exercised at real tag-push. The macOS and Windows packaging
steps and the artifact round trip are the parts of this change that no local
run can prove.

## Version string: checked, nothing derives cross-repo behaviour

Everything version-derived in this workflow is namespaced to this repository,
so an identical version tag elsewhere cannot collide:

- **Asset names** are repo-local and live under this repo's release.
- **Homebrew** URLs are absolute
  (`https://github.com/spelunk-cloud/spelunk/releases/download/v#{version}/…`),
  written into `Formula/spelunk.rb` in the separate `spelunk-cloud/homebrew-spelunk`
  tap, checked out by explicit `repository:`.
- **Scoop** builds every URL from a hardcoded
  `const base = "https://github.com/spelunk-cloud/spelunk"`, including
  `checkver.github` and the `autoupdate` template, into this repo's own
  `bucket/spelunk.json`.
- **Upgrade corpus**: `scripts/upgrade-corpus/generate.sh` pins
  `REPO_SLUG="spelunk-cloud/spelunk"` and refuses any asset not listed in
  `checksums.txt`. The checked-in corpus wings are produced by 0.8.3 / 0.9.2 /
  0.9.3 / 0.9.5, all historical; nothing expects an asset at the tag being
  released. `.github/workflows/upgrade-corpus.yml` fetches one pinned old
  binary at `v0.9.3` from `${GITHUB_REPOSITORY}` and verifies it against a
  pinned sha256.
- **Installers** pin `REPO="spelunk-cloud/spelunk"` and resolve `releases/latest`
  on that repo.

The only behaviour actually derived from the version *string* is the pre-release
predicate (`-rc` / `-beta` / `-alpha`), which gates `--prerelease` and the two
package-manager jobs. It reads this repo's own tag and references nothing
external. No version bump was made here.


usercise (commented):
## Review: PASS

Reviewed independently in my own detached worktree at `9b67f08`, re-deriving each claim by running things rather than reading them. Nothing blocking. Two non-blocking hardening suggestions and two nits at the end.

First, a framing point that matters more than anything else in this diff: **the 15/15 green CI is not evidence for this change.** No job in `ci.yml` lints or executes `release.yml`, and "Release script tests" runs only `node --test .github/scripts/write-deb-control.test.js` — it never touches `scripts/release-dry-run.sh`. The green checks establish that the workspace still compiles on three platforms; they say nothing about the packaging. Everything below is therefore from direct observation.

---

### 1. Release globs — verified, high confidence

This is the claim I spent the most on, since no local run reaches `gh release create`.

The premise the whole thing rests on is that `download-artifact@v8` with no `name` and `merge-multiple: true` downloads **every** artifact in the run and flattens them into one directory. I confirmed that empirically from the v0.9.7 release run's own log (run `31023656916`, "Create GitHub Release" job):

```
Found 5 artifact(s)
No input name, artifact-ids or pattern filtered specified, downloading all artifacts
Starting download of artifact to: /home/runner/work/spelunk/spelunk/artifacts   (x5, same path)
Total of 5 artifact(s) downloaded
```

Five distinctly-named artifacts, all landing at `artifacts/` with no per-artifact subdirectory — and the globs then matched, since v0.9.7 published all five assets. Two consequences follow directly:

- The new artifacts are downloaded because **no filter exists**, not because their names happen to resemble the old ones. Adding an artifact name cannot be missed by a step that takes all of them.
- They land flat, so `artifacts/*.tar.gz` / `artifacts/*.zip` reach them.

I then simulated the flattened directory with all nine expected filenames and expanded the exact glob triple from the workflow. All nine assets are passed to `gh release create`, including all four `spelunk-export-*`:

```
artifacts/spelunk-export-v0.9.8-aarch64-apple-darwin.tar.gz
artifacts/spelunk-export-v0.9.8-aarch64-unknown-linux-gnu.tar.gz
artifacts/spelunk-export-v0.9.8-x86_64-unknown-linux-gnu.tar.gz
artifacts/spelunk-v0.9.8-aarch64-apple-darwin.tar.gz
artifacts/spelunk-v0.9.8-aarch64-unknown-linux-gnu.tar.gz
artifacts/spelunk-v0.9.8-x86_64-unknown-linux-gnu.tar.gz
artifacts/spelunk-export-v0.9.8-x86_64-pc-windows-msvc.zip
artifacts/spelunk-v0.9.8-x86_64-pc-windows-msvc.zip
artifacts/spelunk_0.9.8_amd64.deb
```

All nine basenames are distinct, so `merge-multiple`'s overwrite-on-collision behaviour is not in play. No `nullglob`/`failglob` is set, but every pattern matches, and a tag matching `v*.*.*` contains no glob metacharacters.

**Confidence: high.** The residual risk is not in the glob — it is upstream, in whether the artifact gets created at all. See suggestion (a).

### 2. Product artifact still contains exactly one file — verified

`path: ${{ env.ARCHIVE }}` is unchanged, and `ARCHIVE` is written only by the two pre-existing packaging steps. The new steps write a distinct variable (`EXPORT_ARCHIVE`) and upload under a distinct artifact name, so nothing can leak into the product artifact. The value is an exact filename with no wildcard, so `upload-artifact`'s pattern handling cannot broaden it. Step ordering is fine: both `GITHUB_ENV` writes land before their respective consumers.

`deb` and `tarball-smoke` therefore still download and extract exactly what they did before. `tarball-smoke` downloading two artifacts into the same `path: tarball` is safe — distinct filenames, and `download-artifact` does not clear the destination.

### 3. Windows `Compress-Archive` — verified against the existing step

I ran both steps side by side under pwsh 7.4 and read the resulting zip central directories:

```
=== spelunk-v0.9.8-x86_64-pc-windows-msvc.zip ===
  entry: spelunk.exe
  entry: spelunk-server.exe
=== spelunk-export-v0.9.8-x86_64-pc-windows-msvc.zip ===
  entry: spelunk-export.exe
```

Same flattening (basename at archive root, no `target/.../release/` prefix), same working directory, same naming convention, same `Out-File -Encoding utf8 -Append` idiom for `GITHUB_ENV`. They agree.

I also checked the failure mode: `Compress-Archive` on a missing input throws `InvalidOperationException`, and GitHub sets `$ErrorActionPreference = 'stop'` for `shell: pwsh`, so a missing `spelunk-export.exe` fails the step loudly rather than producing an empty artifact. The unix counterpart fails the same way via `set -euo pipefail` (`tar` exits non-zero on a missing member).

Separately: `spelunk-export.exe` demonstrably builds on Windows MSVC — the `Test (windows-latest)` job runs `cargo nextest run --lib --bins ...` unscoped, so it compiles the bin. Worth noting because the crate landed *after* `v0.9.7`, so no release run has ever built it on Windows or macOS.

### 4. glibc ceiling over three binaries — verified, it really fails

Extracted the loop body verbatim with `objdump` stubbed, and drove the third binary through five cases:

| third binary's symbols | result |
|---|---|
| `GLIBC_2.2.5 GLIBC_2.30` | pass (exit 0) |
| `GLIBC_2.31` | pass (exit 0) |
| `GLIBC_2.2.5 GLIBC_2.34` | **fail**, exit 1, `spelunk-export requires GLIBC_2.34, above the glibc 2.31 floor` |
| `GLIBC_2.4 GLIBC_2.32` | **fail**, exit 1 |
| `GLIBC_PRIVATE` only | **fail**, exit 1, "no valid versioned GLIBC symbol found" |
| none | **fail**, exit 1 |

The loop body runs in the step's own shell, so `exit 1` terminates the step rather than the iteration — it does not loop past a failure or discard the result. The false-pass guards apply to the third binary exactly as to the first two. A missing binary also aborts, since `raw="$(objdump -T "$path")"` propagates under `set -e`.

Note `strip = true` in `[profile.release]` strips `.symtab`, not `.dynsym`, so `objdump -T` still sees the versioned glibc references. Confirmed on the real stripped binary below.

### 5. The smoke test proves the static SQLite link — verified

Built `spelunk-export` for `x86_64-unknown-linux-gnu` inside `debian:11` (amd64) and inspected the stripped artifact:

```
ELF 64-bit LSB pie executable, x86-64, dynamically linked, ..., stripped
NEEDED  libgcc_s.so.1 / libpthread.so.0 / libm.so.6 / libdl.so.2 / libc.so.6
max versioned glibc symbol = GLIBC_2.30
no sqlite dynamic dependency
```

Then ran the packaged binary against a **real store written by a released version** — `crates/spelunk-cli/tests/fixtures/upgrade-corpus/wings/memory-v0.9.5/memory.db` — in all four smoke images (`debian:11`, `ubuntu:20.04`, `debian:12`, `ubuntu:24.04`). None of the four base images carries `libsqlite3` at all. All four produced identical, correct output:

```
inventory: {"exists": true, "bytes": 3788800, "contents":[{"notes":4},{"memory_edges":1}]}
export:    4 memory_entry, 1 supersedes
           the dump reads back as exactly what was read from the store
```

So the smoke test's shape is right: `export` + `grep` is what carries the proof. Worth stating that `inventory` alone would not have been evidence — it reports a missing path as `exists: false` and exits 0, so it would pass on a wrong `--store`. The `export` + `grep -q "tarball smoke"` pair is what makes the assertion real.

I also checked the store-location step in isolation: after the exact `git init` / `spelunk init --no-index` / `spelunk memory add` sequence, exactly one `memory.db` exists, at `<workdir>/.spelunk/memory.db`, so `find /w "$HOME" -name memory.db -print -quit` resolves deterministically to it.

And I verified the dry-run's linkage assertions actually fire rather than being decorative. Against a purpose-built binary linked with `-lsqlite3`, and against `dbus-send`:

```
libsqlite3.so.0 => ...   -> ASSERTION WOULD FIRE: links a system SQLite
libdbus-1.so.3  => ...   -> CAUGHT: links libdbus
```

(Aside: Debian's own `/usr/bin/sqlite3` statically links the amalgamation and is *not* a valid probe for this — worth knowing if anyone re-tests it.)

### 6. Name neutrality — verified

Scanned every added line for destination, product and migration vocabulary. The only hit is the literal binary name `spelunk-export` in the Windows path. Nothing in the workflow, the script, `docs/releasing.md` or the CHANGELOG names a destination, another product, a migration, or an end of life. The CHANGELOG wording ("fetched on its own, verified, run against a store and deleted again") stays behavioural.

### On the dynamic-linkage framing — the author is right

The report's careful wording is accurate, and it matters. `rusqlite`'s `bundled` feature compiles SQLite in, so there is no `libsqlite3` in `NEEDED` (confirmed above by `objdump -p`). But the binary is a perfectly ordinary dynamically-linked PIE against platform glibc — five `NEEDED` entries, `GLIBC_2.30` ceiling — **not** a musl-style fully static artifact. The glibc 2.31 floor therefore applies to it exactly as to the other two, and adding it to the ceiling check is correct rather than theatre. Had it actually been fully static there would be no versioned GLIBC symbols at all, and the check's own false-pass guard would have failed the release with "no valid versioned GLIBC symbol found" — so the check would have told us either way.

---

### Non-blocking suggestions

**(a) `if-no-files-found: error` on both upload steps.** `actions/upload-artifact@v7` defaults this to `warn` (confirmed in the v7 `action.yml`), and nothing in this repo overrides it. The packaging steps do fail loudly on a missing binary, so this is defence in depth rather than a live bug — but the consequence is asymmetric in a way worth closing:

- A missing **product** archive on macOS/Windows is caught downstream: `update-homebrew-formula` and `update-scoop-manifest` read those exact asset names' digests, and `update-homebrew-formula.js` runs `requireValue` + `assertSha256`, which throw on an empty digest.
- A missing **export** archive on macOS or Windows has **no consumer whatsoever**. `tarball-smoke` covers only the two Linux targets. The release would go green having published no helper for those two platforms.

That is precisely the "succeeds while publishing nothing" shape this repo has already produced once. Two lines close it permanently.

**(b) Consider a post-release asset assertion.** There is no post-release verification step in `release.yml` or checklist in `docs/releasing.md`. A short step after `gh release create` asserting the expected asset count (or the four `spelunk-export-*` names) would turn every remaining silent-absence failure into a loud one. Out of scope here; noting it because this PR raises the expected asset count from 5 to 9.

### Nits

- `scripts/release-dry-run.sh` derives `VERSION` from `Cargo.toml`, so it produces `spelunk-export-0.9.7-dryrun-<target>.tar.gz` — no `v` prefix, unlike CI's `github.ref_name`. Consistent with how the `.deb` already behaves, so not a defect, but it means the dry run does not validate the exact published asset name.
- `find /w "$HOME" -name memory.db -print -quit` depends on exactly one `memory.db` existing; `find` traversal order is not sorted. True today (verified), but a direct `/w/.spelunk/memory.db` would fail with a clearer message if that ever changes.

### What I did *not* run, so the next reader knows the coverage

I did **not** run `scripts/release-dry-run.sh` end to end. This host is arm64 and the script pins `--platform linux/amd64`, so the full workspace build with LTO runs under QEMU — hours, for a `.deb` path this PR does not touch. Instead I verified the script statically (`bash -n` clean, `shellcheck -S warning` clean) and executed the new `package_export_archive` stage in isolation against a real container-built binary:

```
=== release-dry-run: package export tool archive + check dynamic linkage ===
libgcc_s.so.1 / libpthread.so.0 / libm.so.6 / libdl.so.2 / libc.so.6
STAGE EXIT=0
target/release-dry-run/spelunk-export-0.9.7-dryrun-x86_64-unknown-linux-gnu.tar.gz
tar -tzf -> spelunk-export        (single flat entry, matching the product tarball's layout)
```

Also still unexercised by anything, here or in CI, and only reachable at real tag-push: the macOS and arm64-Linux packaging legs, the `upload-artifact`/`download-artifact` round trip as a round trip, and the `digest` field on the new assets. The PR description states these limits accurately.

